PR #24, opened 2022-01-01
Add Climb collection

gibboj:
I added climb as a separate collection  I think if we want to have a higher area-less view of things it will work better, but let me know if you prefer having all the info in the area. I didn't switch the areas to just store the climb UIDs yet.


vnugent:
Let's try this new approach if it makes more sense for us to work with climbs in the frontend.  It's my naive assumption that users would tend to browse sibling climbs in the same area.  By placing climbs as subdocuments of a leaf area we don't need to do a traditional sql "join".

Issue #90, opened 2022-07-12
Use a proper logging framework


vnugent:
Use something like https://github.com/pinojs/pino for logging

PR #92, opened 2022-07-19
Added pino for logging OpenBeta/openbeta-graphql#90

admanny:


admanny:
Additionally we can integrate https://github.com/pinojs/pino-pretty to format the logs in a more visually appearing way - @vnugent

vnugent:
we're currently running Mongo on our own Kubernetes. Do you have suggestions for a log storage solution?

admanny:
I've used logstash->elasticsearch in the past. But for this use-case I believe a Mongo collection should suffice

vnugent:
does it mean we're pushing the log to the same climb DB?

I'm currently using Logflare free plan to ingest frontend log from Vercel (for occasional troubleshooting).

Could their [BYOB](https://logflare.app/pricing) plan work for us?

vnugent:
@all-contributors add @admanny for code

PR #108, opened 2022-08-23
Area search

CocoisBuggy:
I've generalized some of the code for adding new collections. Everything is importing to typescript without errors, I'm not sure how to properly test it though...

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,9 @@
     "refresh-db": "./refresh-db.sh",
     "seed-usa": "yarn build && node build/db/import/usa/USADay0Seed.js",
     "init-db": "yarn build && node build/db/utils/jobs/UpdateAllRunner.js",
-    "update-search": "tsc ; node build/db/export/Typesense/Typesense.js"
+    "update-climb-search": "tsc ; node build/db/export/Typesense/Typesense.js --climbs",
+    "update-area-search": "tsc ; node build/db/export/Typesense/Typesense.js --area",
+    "update-search": "tsc ; node build/db/export/Typesense/Typesense.js --areas --climbs"
   },
   "standard": {
     "plugins": [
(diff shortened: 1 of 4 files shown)


vnugent (changes requested):
Thanks for this!   

In current version we accumulate the chunk array to max `chunkSize` and them push to Typesense **as** we loop through the Mongo result set.  Therefore at any given time there are only at most 5K (chunkSize) documents in memory.  

In the PR we add the entire Mongo collection to `chunks` array, which can be very large (~125K climb documents).

package.json
```diff
@@ -55,7 +55,7 @@
     "refresh-db": "./refresh-db.sh",
     "seed-usa": "yarn build && node build/db/import/usa/USADay0Seed.js",
     "init-db": "yarn build && node build/db/utils/jobs/UpdateAllRunner.js",
-    "update-search": "node build/db/export/Typesense/Typesense.js"
+    "update-search": "tsc ; node build/db/export/Typesense/Typesense.js"
```
vnugent: can we have a new and separate job?  Ex:  `yarn update-search-climbs` and `yarn update-search-areas`?  This way we can test and run each job independently without impacting one another.

vnugent (changes requested):

CocoisBuggy:
> Thanks for this!
> 
> In current version we accumulate the chunk array to max `chunkSize` and them push to Typesense **as** we loop through the Mongo result set. Therefore at any given time there are only at most 5K (chunkSize) documents in memory.
> 
> In the PR we add the entire Mongo collection to `chunks` array, which can be very large (~125K climb documents).

Ah, of course. I'll refactor for better memory use

package.json
CocoisBuggy: Sure, I'll add that in

vnugent:
@all-contributors add @CocoisBuggy for code and ideas

PR #135, opened 2022-09-11
Update README for troubleshooting steps

DarrenZLew:
- Moves troubleshooting steps into their own section
- Adds a troubleshooting step related to the owner of the keyfile not being the same owner as the mongodb process


vnugent:
@all-contributors add @DarrenZLew for ideas and code

Issue #157, opened 2022-09-19
Create a grade context (option) at the country level


vnugent:

DarrenZLew:
Created a PR https://github.com/OpenBeta/openbeta-graphql/pull/163 to add a grade context field for each area. Defaults to the country level.

Issue #141, opened 2022-09-13
Improve crag stats aggregation to support new grade schema


vnugent:
Related #111 

The statistics aggregate function uses `yds` field.   We need to migrate to the new grade schema..

https://github.com/OpenBeta/openbeta-graphql/blob/staging/src/db/utils/Aggregate.ts#L61

### Problem

How do we know which field to use? French/YDS?

DarrenZLew:
In terms of the grading system, I think it would make the most sense to default to what is most commonly accepted for that crag/area/country. So for our current USA data and supporting grade types, this would be YDS and V-scale.

An example implementation would be https://www.thecrag.com/en/article/gradesonthecrag#article.gradesonthecrag.header.grade-contexts. They default to what is accepted at the country level, but this can be overwritten at individual crags/areas within the country.

vnugent:
I created https://github.com/OpenBeta/openbeta-graphql/issues/157 to track this new grade context.

DarrenZLew:
Created a PR to help update this: https://github.com/OpenBeta/openbeta-graphql/pull/163

Some additional work needed to update the statistics function to move some logic into sandbag

Issue #140, opened 2022-09-13
Improve USA seed scripts to support new Grade schema


vnugent:
Related #111 

The climb data seed script (loading USA data from JSON files to db) writes to field `YDS`.

https://github.com/OpenBeta/openbeta-graphql/blob/staging/src/db/import/ClimbTransformer.ts#L18

### To do:
- [ ] Use new grade field

DarrenZLew:
Created PR https://github.com/OpenBeta/openbeta-graphql/pull/163 to start using the ~~grade~~ grades field

vnugent:
@DarrenZLew is this done?

DarrenZLew:
I believe so, yes.

Issue #111, opened 2022-08-25
Support French and other grading systems


vnugent:
The current schema is hard-coded to YDS.  
- [x] Update the Climb schema to support other systems (we can prioritize the French system for now)
- [x] Update GQL queries
- [x] Migrate frontend to new fields (PR TBD)
- [x] Update USA import script https://github.com/OpenBeta/openbeta-graphql/issues/140
- [x] Update statistics script https://github.com/OpenBeta/openbeta-graphql/issues/141
- [ ] Delete old yds fields from schema and GQL

vnugent:
suggestion
>  Considering that conversions are subjective, and might change in the future, it might be worth it to store the original grade (not the universal representation), so new conversion scales can properly change with it

PR #163, opened 2022-09-26
Add grade context to area and climb schemas

DarrenZLew:
Overall changes for completing the tasks below to help complete the larger scoped #111.

#157 Create grade contexts defaulted to country level. There is a cascading effect of sub-areas using the parent area's grade context if no grade context is imported into that specified area. This continues to the crag level. Climbs of the crag will inherit that crag's grade context (currently not allowed to import at the climb level).

#140 Add new grades schema that maps to grade context. Updates the area and climb schemas to use the new grades field

#141 Update statistics aggregate function to use new grades field.

---------------------
Noticed some data integrity issues:
- Will we be handling "snow"/"ice" climb types or are these considered "alpine"?
- One climb in the current data set didn't return a climb type. The expectation would be for all climbs to have a climb type.

Todos:
- Moving the getBand/score logic out of openbeta-graphql into sandbag repo, so there's a single source of truth.
- Moving grade scales out of openbeta-graphql into sandbag repo, so there's a single source of truth.
- Add more grade contexts beyond US and FR.
- Add functionality to update the grade context for an area. Should this only be allowed by an admin?

### DIFF
--- a/src/db/AreaSchema.ts
+++ b/src/db/AreaSchema.ts
@@ -4,6 +4,7 @@ import muuid from 'uuid-mongodb'
 import { AreaType, IAreaContent, IAreaMetadata, AggregateType, CountByGroupType, CountByDisciplineType, CountByGradeBandType, DisciplineStatsType, OperationType } from './AreaTypes.js'
 import { PointSchema } from './ClimbSchema.js'
 import { ChangeRecordMetadataType } from './ChangeLogType.js'
+import { GradeContexts } from '../grade-utils.js'
 
 const { Schema, connection } = mongoose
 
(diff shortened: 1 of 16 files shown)


vnugent:
Thank you for this big PR!

> Add functionality to update the grade context for an area. Should this only be allowed by an admin?

I created #165 for this task.  If we piggyback context as an updatable field in the area mutation then the permission check is already taken care of.  FYI currently we don't have fine grained [access control](https://github.com/OpenBeta/openbeta-graphql/blob/cad0ccbfb125b2b7fa1c7cd3b19a9187a74385dc/src/auth/permissions.ts). Anyone with `editor` role can perform a create/update/delete.

> Will we be handling "snow"/"ice" climb types or are these considered "alpine"?

Would the `mixed` type be acceptable for waterfall ice and mixed alpine ice + rock?

Issue #150, opened 2022-09-15
Add yarn lint to precommit hook


vnugent:
we should enable `yarn lint` in git pre-commit hook to catch linting errors.

DarrenZLew:
Added in https://github.com/OpenBeta/openbeta-graphql/pull/169

PR #169, opened 2022-09-27
Add precommit hook

DarrenZLew:
#150 Added husky and pre-commit hook to repo.

Will run the following prior to committing:
yarn lint
yarn test

This is working locally for me, but please verify it works for you as well.


vnugent:
Thank you!
On my Macbook it complained `.husky/precommit` file is not executable.  I've pushed that fix to your PR.

PR #203, opened 2022-12-08
Coco/code docs

CocoisBuggy:
Continuing the work of documentation + minor refactoring of schema definition/validation flow.

## Docs

There is still a lot of work to do with properly documenting data intention and scopes, but it's important to do it as it will improve new developer on-boarding and comprehension of how data flows through the API layer. There are still a handful of methods and flows that are in desperate need of extra documentation, but I am stalling work on those to try and get some feedback on the second part of the PR

## Schema Def

I want to get some feedback on this concept before I proceed, the Idea is that schema definition will take place in files with extensions of `.gql` or `.graphql` which VSCode understands as being graphql schema files. In my opinion this communicates file intent more clearly and makes the directory much easier to parse as the files now indicate their contents with their icons / extensions (I also use exa with icons in my terminal shell, so `ls` also shows their icons). There is the added benefit that vscode can better communicate trivia schema errors with the extended syntax hilighting.

Apollo will still load and validate the files in an identical way, the only difference is that the strings are read from a file rather than defined in string literals.

The final thing is that services like GitHub will now provide syntax hilighting in browser [click here to see](https://github.com/OpenBeta/openbeta-graphql/blob/coco/codeDocs/src/graphql/schema/Area.gql)


vnugent:
I'm near done with part 1 of #201.  Can you please wait and rebase after I merge #201?  it touches a lot of the same files.

CocoisBuggy:
Not a problem. 90% cosmetic changes so I'll rebase after merge of #201

vnugent:
#201 merged.  Can you please rebase?

CocoisBuggy:
@vnugent Should be good to go. Migrated the GQL spelling changes / deprecation notices + added a schema file for the new climb mutations

vnugent:
See screenshot:

Screen Shot 2022-12-10 at 2 10 01 PM

CocoisBuggy:
Weird, not sure what the correct procedure to that would have been.
Strictly speaking I thought the merge conflicts would have been solved, but I didn't properly signal to git HOW they were solved.

Looks to be fixed now

PR #214, opened 2023-01-05
Add x media schema and tag schema

siman4457:
## XMedia
`XMedia` stands for "Extended Media" and consists of any type of media that a user uploads with a Post. Currently we only support uploading photos with a Post. Eventually, XMedia should eventually replace the current `Media` collection.

### Workflow of creating a Post with XMedia (WIP):
1) User selects a photo, adds description and tag(s)
2) Photo gets uploaded to Sirv on the frontend
3) Get the response from Sirv
4) Trigger addPost API with Sirv response and user's input (description and tags)

### New Queries and Mutations
Queries:
- getXMedia

Mutations:
- addXMedia
- removeXMedia

## Tags
`Tags` link XMedia to climb(s) or area(s). XMedia will contain an array of tagIds.

### New Queries and Mutations
Queries:
- getTags

Mutations:
- addTag
- deleteTag

### DIFF
--- a/src/db/XMediaSchema.ts
+++ b/src/db/XMediaSchema.ts
@@ -1,0 +1,47 @@
+import mongoose from 'mongoose'
+import mongooseLeanVirtuals from 'mongoose-lean-virtuals'
+import muuid from 'uuid-mongodb'
+
+import { XMediaType } from './XMediaTypes.js'
+
+const { Schema } = mongoose
+
+const XMediaSchema = new Schema<XMediaType>({
+  userId: {
+    type: Schema.Types.ObjectId,
+    required: true
+  },
+  mediaType: {
+    type: Schema.Types.Number,
+    required: true
+  },
+  mediaUrl: {
+    type: Schema.Types.String,
+    required: true
+  },
+  mediaUuid: {
+    type: 'object',
+    value: { type: 'Buffer' },
+    default: () => muuid.v4(),
+    required: true,
+    unique: false,
+    index: true
+  },
+  tagIds: {
+    type: [Schema.Types.ObjectId],
+    required: true
+  }
+}, {
+  _id: true,
+  strictQuery: 'throw',
+  toObject: {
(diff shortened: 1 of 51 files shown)


src/db/XMediaSchema.ts
```diff
@@ -0,0 +1,47 @@
+import mongoose from 'mongoose'
+import mongooseLeanVirtuals from 'mongoose-lean-virtuals'
+import muuid from 'uuid-mongodb'
+
+import { XMediaType } from './XMediaTypes.js'
+
+const { Schema } = mongoose
+
+const XMediaSchema = new Schema<XMediaType>({
+  userId: {
```
vnugent, line 10: User id in Auth0 is in uuid v4 format (Ex: `96373086-2d70-4e92-8036-3947d4cec3f9`) so we'd need to use MUUID here.  See `mediaUuid`  below for declaration.

vnugent (changes requested):

vnugent:
@siman4457 Looking great.  I requested a small change to userId then we're good to merge.

src/db/XMediaSchema.ts
siman4457: I've made this change and also changed destinationId on XMedia Schema to also use MUUID since this is what Media Schema is using. Lastly, I added onModel to XMedia (also copied from Media Schema).

src/graphql/posts/mutations.ts
```diff
@@ -21,7 +21,6 @@ const PostMutations = {
   removePost: async (_: any, { input }: {input: RemovePostInputType}) => {
     const PostModel = getPostModel()
```
vnugent, line 22: just in case you haven't planned to,  it's a good idea to put database logic in its own Data source classes so that the code in here is light.

siman4457: I'm planning to do that in the future. I wanted to start out by keeping it simple.

PR #217, opened 2023-01-14
feat(export): add generic json exporter

Silthus:
This PR adds a generic JSON exporter for areas and climbs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,9 @@ dist
 build
 tmp
 .env.local
+
+# Intellij
+.idea/
+*.iml
+
+/export/
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -11,5 +11,6 @@ services:
       ME_CONFIG_MONGODB_URL: mongodb://${MONGO_INITDB_ROOT_USERNAME}:${MONGO_INITDB_ROOT_PASSWORD}@mongo_opentacos:27017/opentacos?authSource=admin&replicaSet=rs0
     depends_on:
     - mongo_opentacos
+    - mongosetup
 volumes:
   opentacos_mongodb_data:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,24 @@ services:
       test: test $$(echo "rs.initiate().ok || rs.status().ok" | mongo -u $${MONGO_INITDB_ROOT_USERNAME} -p $${MONGO_INITDB_ROOT_PASSWORD} --quiet) -eq 1
       interval: 10s
       start_period: 30s
-    command: ["--replSet", "rs0", "--bind_ip_all", "--keyFile", "/opt/keyfile/keyfile"]
+    entrypoint:
+      - bash
+      - -c
+      - |
+        cp /opt/keyfile/keyfile /data/db/replica.key
+        chmod 400 /data/db/replica.key
+        chown 999:999 /data/db/replica.key
+        exec docker-entrypoint.sh $$@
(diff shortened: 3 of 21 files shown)


vnugent:
@Silthus Just checking to see if you need anything from me?

Silthus:
> @Silthus Just checking to see if you need anything from me?

Hey @vnugent,
yes I am still working on this, just have to find the time for it :)

Just to summarize from our conversation in discord:

- I will create a generic JSON export function 
- the function exports all entries in the database to JSON files
- one file per area document

An idea I had, but need your opinion on: how about doing an hierarchical export and place the climbs within the leaf documents?

```
export/usa/utah/some-area/some-sub-area.area.json
export/usa/utah/some-area.area.json
export/usa/utah/utah.area.json
```

vnugent:
> export/usa/utah/some-area/some-sub-area.area.json

Either that or using a flat structure and adding a delimited `path` field. Ex: `path: usa | utah | some-area...`

FYI our old data repo (before we adopted mongo): https://github.com/OpenBeta/opentacos-content

Silthus:
I had quite some trouble setting up the local dev environment so I did some changes that get rid of the keyring problem and should work ouf of the box.

https://github.com/OpenBeta/openbeta-graphql/commit/8ce788207e73b3b1fdfc51d1c8f6dba6438ce662

Is it ok if I include these changes in this PR or should I split them?

vnugent:
What issues were you having?  Did setting the keyfile permission alone as mentioned in the FAQ fix it?
My only concern is introducing a bash script would make life a bit harder for folks on Windows. 

I don't use Windows but I think you'll need to enable/install WSL or something like that?

Silthus:
> What issues were you having?  Did setting the keyfile permission alone as mentioned in the FAQ fix it?
> 
> My only concern is introducing a bash script would make life a bit harder for folks on Windows. 
> 
> 
> 
> I don't use Windows but I think you'll need to enable/install WSL or something like that?

Unfortunately the FAQ didn't help as chmod are Linux Commands and don't exist in Windows. And running them through WSL doesn't work either.

The script I added solely gets executed inside the docker container to setup the replica set. I moved the setting of the permissions inside the docker container so the FAQ should be obsolete.

Also WSL or some kind of Unix system is required anyways because the `refresh-db` command executes a shell script. 

It might be helpful to streamline the local development and move more of these steps inside the container to also work on Windows without WSL.

Another problem I noticed is that module importing inside Typescript files does not seem to work correctly as I can't import like so:

```ts
import { Something} from "./somewhere"
```

I need to add a `.js` extension which is weird, because it is a typescript file.

```ts
import { Something} from "./somewhere.js"
```

Barrel imports don't seem to work either.

vnugent:
You're right.  If it's too much trouble to split you can submit the PR as is.  

> import { Something} from "./somewhere"

Yes it's a known problem that I've never got around to fix.  If you can address it that'd be awesome (same PR is fine if it's a simple change).  Thanks.

Silthus:
Ok this PR should be ready now.
I included the local dev setup changes in a separate commit.
There are tests for the logic, but they fail in the verify step due to the `*.js` import stuff and pass when I remove the `.js` extensions from the import. But then the execution fails so I readded them.

I will take a look at the Javascript imports in a separate PR when I find the time.

To use this, a cronjob would need to be setup with a wrapper script that executes the export and then commits and pushes the changes to a git repo.

```sh
yarn export:json:full
```

I will add a timestamp parameter in a future pr to only dump from a certain point in time.

vnugent:
@all-contributors add @Silthus for ideas & code

PR #220, opened 2023-02-04
feat(export): add export script for cron job

Silthus:


vnugent:
Would you be able to help creating a K8s cron definition as well?

PR #251, opened 2023-03-19
(obsolete) feat: GradeType "UIAA" and datafield "bolts"

l4u532:
OpenBeta misses some database fields/types to cover the needs for climbs in Central Europe (Germany, Austria, Switzerland, etc.). For one, the UIAA grading system. For the other, the number of bolts (fixed anchors) for a climb, as sport climbs are the vast majority, trad climbs are more of a niche.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,10 +114,7 @@ tmp
 
 # VSCode
 *.code-workspace
- .vscode/settings.json
 
 # DB Export
 /export/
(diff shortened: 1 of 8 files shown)


l4u532:
@vnugent could you be so kind and delete this PR? It exposes my full name, as I forgot to change the user.name, which I would like to avoid.

PR #252, opened 2023-03-19
add UIAA gradetype

l4u532:
OpenBeta misses some database fields/types to cover the needs for climbs in Central Europe (Germany, Austria, Switzerland, etc.). For one, the UIAA grading system. For the other, the number of bolts (fixed anchors) for a climb, as sport climbs are the vast majority, trad climbs are more of a niche.

### DIFF
--- a/src/GradeUtils.ts
+++ b/src/GradeUtils.ts
@@ -90,11 +90,12 @@ export const gradeContextToGradeScales: Partial<Record<GradeContexts, ClimbGrade
     sport: GradeScales.UIAA,
     bouldering: GradeScales.FONT,
     tr: GradeScales.UIAA,
+    deepwatersolo: GradeScales.FRENCH,
     alpine: GradeScales.UIAA,
     mixed: GradeScales.UIAA, // TODO: change to MI scale, once added
     aid: GradeScales.UIAA,
     snow: GradeScales.UIAA, // TODO: remove `snow` since it duplicates `ice`
-    ice: GradeScales.UIAA // TODO: change to WI scale, once added
+    ice: GradeScales.WI
   }
 }
 
--- a/src/GradeUtils.ts
+++ b/src/GradeUtils.ts
@@ -84,6 +84,17 @@ export const gradeContextToGradeScales: Partial<Record<GradeContexts, ClimbGrade
     aid: GradeScales.AID,
     snow: GradeScales.FRENCH, // SA does not have a whole lot of snow
     ice: GradeScales.WI
+  },
+  [GradeContexts.UIAA]: {
+    trad: GradeScales.UIAA,
+    sport: GradeScales.UIAA,
+    bouldering: GradeScales.FONT,
+    tr: GradeScales.UIAA,
+    alpine: GradeScales.UIAA,
+    mixed: GradeScales.UIAA, // TODO: change to MI scale, once added
+    aid: GradeScales.UIAA,
+    snow: GradeScales.UIAA, // TODO: remove `snow` since it duplicates `ice`
+    ice: GradeScales.UIAA // TODO: change to WI scale, once added
   }
 }
 
--- a/src/graphql/schema/Climb.gql
+++ b/src/graphql/schema/Climb.gql
@@ -156,7 +156,7 @@ type GradeType {
   font: String
   """
   UIAA grading system, typically used in Central Europe (e.g. Germany, Austria, Switzerland).
-  Uses Roman numerals, e.g. "7-", "7", "7+". (Arabic numerals, like "VII-", are not supported).
+  Uses Arabic numerals, e.g. "7-", "7", "7+". (Roman numerals, like "VII-", are not supported).
   https://en.wikipedia.org/wiki/Grade_(climbing)#UIAA
   """
   uiaa: String
--- a/src/graphql/schema/Climb.gql
+++ b/src/graphql/schema/Climb.gql
@@ -154,6 +154,12 @@ type GradeType {
   https://www.99boulders.com/bouldering-grades#font-scale-aka-fontainebleau-scale
   """
   font: String
+  """
+  UIAA grading system, typically used in Central Europe (e.g. Germany, Austria, Switzerland).
+  Uses Roman numerals, e.g. "7-", "7", "7+". (Arabic numerals, like "VII-", are not supported).
+  https://en.wikipedia.org/wiki/Grade_(climbing)#UIAA
+  """
+  uiaa: String
 }
 
 """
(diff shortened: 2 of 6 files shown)


l4u532:
- [x] Merge adding UIAA to sandbag https://github.com/OpenBeta/sandbag/pull/81

vnugent:
@l4u532 thanks for the PR.  May I ask that you create 2 separate PRs, one for UIAA and one for bolt count field?

For the bolt count, maybe consider using an array to account for multipitch climbs?

bradleyDean:
> @l4u532 thanks for the PR. May I ask that you create 2 separate PRs, one for UIAA and one for bolt count field?
> 
> For the bolt count, maybe consider using an array to account for multipitch climbs?

How are we dealing with variants of routes? Many multipitch routes have slight variations, where you can break off from the standard route and maybe bypass a difficult section, for example, by climbing a bolt ladder or an easier or harder pitch. 
As far as the bolt-count problem goes, how should we handle this?

This really introduces some complexity. Do we already have a method for representing route variants? It seems to me that an array with array index mapping to bolt count per pitch would not easily extend to representing bolt counts for routes with variations. On the other hand, if every variation has its own unique record and name, then there is no problem. 
Sorry I am not yet very familiar with the Schema we use for representing multipitch routes.

l4u532:
@musoke @bradleyDean (1) I suggest creating individual routes for when breaking off of the original route (ie. taking a variant). How else would you be able to tick a route.

(2) For "boltsCount", I agree a simple String array should be the most elegant and user-friendly implementation.

As @musoke suggested, grade should then become a String array, to map difficulty to pitch.

l4u532:
@musoke @vnugent This adds UIAA grade. Removed `boltsCount` from this PR (will add in subsequent PR, then start on #266).

musoke (changes requested):
Could you add some tests?  https://github.com/OpenBeta/openbeta-graphql/pull/255/files has analogous ones for Ewbank grades.

musoke:
Do you want to define a context that uses UIAA too?  https://github.com/OpenBeta/openbeta-graphql/blob/dab9388193d4733e577c454a31354245253fdced/src/GradeUtils.ts#L28 exists but doesn't have its grades defined.

l4u532:
> Do you want to define a context that uses UIAA too?
> 
> https://github.com/OpenBeta/openbeta-graphql/blob/dab9388193d4733e577c454a31354245253fdced/src/GradeUtils.ts#L28
> 
> exists but doesn't have its grades defined.

Good spot. Hm, I don't think I want to define it... [COUNTRIES_DEFAULT_NON_US_GRADE_CONTEXT](https://github.com/l4u532/openbeta-graphql/blob/a2835bc724c3e23e0331f1fdfde6d818217d0dd9/src/GradeUtils.ts#L106) expects this, with a correct-looking mapping. Yet, in Central Europe, where UIAA is dominant, people still use a mix of UIAA and French; in my view, one cannot really definitely pinpoint UIAA to a regional context (apart from alpine, maybe). Besides, `gradeContext` seems to need an overhaul anyway:

[export const gradeContextToGradeScales: Partial<Record<GradeContexts, ClimbGradeContextType>> = {](https://github.com/l4u532/openbeta-graphql/blob/a2835bc724c3e23e0331f1fdfde6d818217d0dd9/src/GradeUtils.ts#L39-L51)

`ice`, `mixed` are denominated in WI and MI grades. (Which are missing grades right now).

Do you think it's feasible to leave the definition out for now?

musoke:
> Yet, in Central Europe, where UIAA is dominant, people still use a mix of UIAA and French; in my view, one cannot really definitely pinpoint UIAA to a regional context (apart from alpine, maybe).

Yes, we need more flexibility there.


> `ice`, `mixed` are denominated in WI and MI grades. (Which are missing grades right now).

I'm working on that, see https://github.com/OpenBeta/openbeta-graphql/pull/283.  Likewise for aid.


> Do you think it's feasible to leave the definition out for now?

Yes.  The main advantage I see is it being an easy way to get some tests in.

src/graphql/schema/Climb.gql
```diff
@@ -152,6 +152,12 @@ type GradeType {
   https://www.99boulders.com/bouldering-grades#font-scale-aka-fontainebleau-scale
   """
   font: String
+  """
+  UIAA grading system, typically used in Central Europe (e.g. Germany, Austria, Switzerland).
+  Uses Roman numerals, e.g. "7-", "7", "7+". (Arabic numerals, like "VII-", are not supported).
```
musoke: Are Roman and Arabic backwards here?

musoke (changes requested):

musoke:
Looks good!  I left one more small query.  Do you want to merge this before or after the ice grades in https://github.com/OpenBeta/openbeta-graphql/pull/283?   If it's easier for you, I  can rebase my PR on this and change the ice grade to WI there.

Could you open an issue describing the problems with the current UIAA context?  I think there are similar questions to be addressed with other contexts - e.g. we don't have a way of allowing both winter ice and alpine ice grades, or aid and aid with mandatory free.  Addressing these may require reworking the grade context model significantly.

src/graphql/schema/Climb.gql
l4u532: Ugh. Fixed.

l4u532:
> Looks good! I left one more small query. Do you want to merge this before or after the ice grades in #283? If it's easier for you, I can rebase my PR on this and change the ice grade to WI there.
> 

I don't mind waiting. Just ping/mention me once you are done :)

> Could you open an issue describing the problems with the current UIAA context? I think there are similar questions to be addressed with other contexts - e.g. we don't have a way of allowing both winter ice and alpine ice grades, or aid and aid with mandatory free. Addressing these may require reworking the grade context model significantly.

I must admit I have not fully understood the motivation behind and all the interconnections of the current gradeContext model. It def needs a re-work, though. I'll have a look.

musoke:
Ok, will ping you when that's done :smile:

musoke (approved):

src/GradeUtils.ts
```diff
@@ -80,6 +80,17 @@ export const gradeContextToGradeScales: Partial<Record<GradeContexts, ClimbGrade
...
+  [GradeContexts.UIAA]: {
+    trad: GradeScales.UIAA,
+    sport: GradeScales.UIAA,
+    bouldering: GradeScales.FONT,
+    tr: GradeScales.UIAA,
+    alpine: GradeScales.UIAA,
+    mixed: GradeScales.UIAA, // TODO: change to MI scale, once added
+    aid: GradeScales.UIAA,
+    snow: GradeScales.UIAA, // TODO: remove `snow` since it duplicates `ice`
+    ice: GradeScales.UIAA // TODO: change to WI scale, once added
```
musoke: The WI scale exists now, please rebase on develop.

musoke (changes requested):

musoke (approved):

musoke:
@l4u532: sorry, I somehow closed this while trying to rebase your changes on develop.  If you either `git push -f` or merge https://github.com/l4u532/openbeta-graphql/pull/1, we'll be set.
Sorry!

musoke:
@all-contributors please add @l4u532  for code and ideas

musoke:
thank you @l4u532!

PR #254, opened 2023-03-27
Create Organization model, schema and types

zichongkao:
### 1. Adds integration tests
These tests make HTTP requests to the GraphQL server, which checks the entire flow particularly the GraphQL parsing layer. Previously we were testing starting one layer below in the Mongoose datasources. 

These integration tests are backed by an in-memory Mongo replset which supports the Mongo sessions and transactions that we use extensively. Previously, the standalone in-memory Mongo server did not support this.

### 2. MUUID GraphQL scalars
A big source of confusion for me was understanding what MUUIDs, UUIDs and IDs are and when we use them. My proposal now is to use MUUIDs across the board except when they are the Mongoose model `_id`, which shouldn't be publicly accessible and aren't used widely anyway. IDs are a default GraphQL scalar type for representing identifiers. So we can use IDs at the GraphQL layer if we need (but we shouldn't!) to expose Mongoose model `_id`s. 

However, for the public-facing ids, eg `area_id`, `orgId`, etc, we use MUUIDs, which are Mongo's primitive UUID construct. As with most UUIDs, these are stored in the backend in binary, but when we handle them in the frontend and make requests with them, they need to be strings. There are two ways to serialize these as strings (which is very confusing) -- one is the "relaxed" mode where they turn into hex strings like "59f1d95a-627d-4b8c-91b9-389c7424cb54" and another is the "canonical" mode where they turn into base64 strings like "WfHZWmJ9S4yRuTicdCTLVA==". It seems we prefer to use "relax" so I've made that consistent everywhere.

This PR completes (there were already code fragments lying around) the creation of a new GraphQL MUUID scalar type. This allows us to define a fixed way to serialize and deserialize the binary. Just by declaring in the GraphQL mutations and queries that something is a MUUID scalar type instead of ID scalar type, the user no longer has to parse strings in the Mongoose data sources (GraphQL automatically deserializes them to binary) and they no longer have to guess what format they'll output in the response as (GraphQL automatically serializes them to relaxed format).

tldr: Use MUUID GraphQL scalar types wherever possible.

### 3. Organization model, schema and types
This was the meat of the PR. From the top of the stack to the bottom, this PR implements:
- The GraphQL queries and resolvers (`/graphql/`)
- The Mongoose datasources (`/model/`)
- The Mongo model and types (`/db/`)

The endpoints so far are [format: `<graphql query>` (ie \<mongoose datasource\>)]
- `query organization` (ie findOrganizationByOrgId)
- `query organizations` (ie findOrganizationsByFilter)
- `mutation addOrganization` (ie addOrganization)
- `mutation updateOrganization` (ie updateOrganization)

## Followup tasks
- [ ] Add `mutation deleteOrganization`
- [ ] Add basecamp flow to create/delete organizations.
- [x] Add integration test for `query organizations`. This will require testing `associatedAreaIds` and `excludedAreaIds`.
- [x] Add integration test for `_changes`. 
- [ ] Add permissioning for updateOrganization. Only org_admins that administer the org should be allowed to call it, not all org_admins. Currently there are no checks for calling updateOrganization.
- [ ] In order to complete point 5, we first need to update the user model to store associated organizations. Need API calls to read and update this field. 
- [ ] Once 6 and 5 are done, update integration test to check that only the right org_admins can update the org.


src/model/MutableOrganizationDataSource.ts
```diff
@@ -0,0 +1,70 @@
...
+    return rs1[0].toObject()
+  }
+}
+
+export const newOrganizationHelper = (displayName: string, orgType: OrgType): OrganizationType => {
+  const _id = new mongoose.Types.ObjectId()
+  const orgId = muuid.v4()
+  return {
+    _id,
+    org_id: orgId,
```
vnugent: please use camelCase? :)

src/db/OrganizationSchema.ts
```diff
@@ -0,0 +1,70 @@
...
+const { Schema, connection } = mongoose
+
+const ChangeRecordMetadata = new Schema<ChangeRecordMetadataType>({
+  user: {
+    type: 'object',
+    value: { type: 'Buffer' },
+    required: true
+  },
+  historyId: { type: Schema.Types.ObjectId, ref: 'change_logs' },
+  prevHistoryId: { type: Schema.Types.ObjectId, ref: 'change_logs' },
```
vnugent: I thinking about retiring this field.  Initially I thought it'd be useful to have both current and previous history id so we can build a simple query to show previous change. However, when doing bulk updates (using bulkWrite) it's difficult to get the current historyId to assign to prevHistoryId.

vnugent (commented):
see feedback inline

src/db/OrganizationSchema.ts
```diff
@@ -0,0 +1,69 @@
...
+  historyId: { type: Schema.Types.ObjectId, ref: 'change_logs' },
+  operation: {
+    type: Schema.Types.Mixed,
+    enum: Object.values(OperationType),
+    required: true
+  },
+  seq: { type: Number, required: true, default: 0 }
+}, { _id: false, timestamps: false })
+
+const ContentSchema = new Schema<IOrganizationContent>({
```
zichongkao, line 24: Not sure if splitting out to a sub-schema has any value. But we do it quite liberally in the AreaSchema and elsewhere. Eg. `area.metadata`, `area.content`. I would be interested to understand the rationale for when to break things out and when not to!

src/db/OrganizationSchema.ts
```diff
@@ -0,0 +1,69 @@
...
+    unique: true,
+    index: true
+  },
+  displayName: { type: String, required: true, index: true },
+  orgType: {
+    type: Schema.Types.Mixed,
+    enum: Object.values(OrgType),
+    required: true
+  },
+  associatedAreaIds: [{ type: 'Buffer' }],
```
zichongkao, line 47: Decided to use the `area_id`s which are MUUIDs here, instead of `area._id` which are Mongo `ObjectId`s. After all, both are indexed, so should be roughly as fast as each other (Maybe Mongo has special optimizations for their own native _ids?) But the big upside is that `area_id`s are externally visible, so we can pass them in directly and filter directly rather than having to pass them in and translate into `area._ids`.

src/db/OrganizationTypes.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Organizations are OpenBeta accounts that are owned by organizations
+ * instead of people. For instance, the Boulder Climbing Community might own
+ * one such account. Organizations may have one or more 'org_admins' who
+ * manage the account. They may update organizations' profiles and perform
+ * special actions under the auspices of the organization. For example, org_admins
+ * of local climbing organizations (LCO) may put up bird nesting notices on
+ * the area pages associated with the LCO.
+ * */
+export interface OrganizationType {
```
zichongkao, line 15: I'm actually not sure when something should be a type and when it should be an interface. What's the diff?

src/db/OrganizationTypes.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Organizations are OpenBeta accounts that are owned by organizations
+ * instead of people. For instance, the Boulder Climbing Community might own
+ * one such account. Organizations may have one or more 'org_admins' who
+ * manage the account. They may update organizations' profiles and perform
+ * special actions under the auspices of the organization. For example, org_admins
+ * of local climbing organizations (LCO) may put up bird nesting notices on
+ * the area pages associated with the LCO.
+ * */
+export type OrganizationType = {
```
zichongkao: Actually not sure when to make this a type vs interface. What's the difference?

src/model/MutableOrganizationDataSource.ts
```diff
@@ -0,0 +1,158 @@
...
+        if (missingAreaIds.length > 0) throw new Error(`Organization update error. Reason: Associated areas not found: ${missingAreaIds.map(m => m.toUUID().toString()).toString()}`)
+        org.set({ associatedAreaIds: associatedAreaIds })
+      }
+      if (excludedAreaIds.length > 0) {
+        const missingAreaIds = await findNonexistantAreas(excludedAreaIds)
+        if (missingAreaIds.length > 0) throw new Error(`Organization update error. Reason: Excluded areas not found: ${missingAreaIds.map(m => m.toUUID().toString()).toString()}`)
+        org.set({ excludedAreaIds: excludedAreaIds })
+      }
+      if (displayName != null) { org.set({ displayName: sanitizeStrict(displayName) }) }
+      if (website != null) { org.set({ 'content.website': sanitizeStrict(website) }) }
```
zichongkao, line 87: We could do alot more to validate that this is a well-formed url. Same for emails below. But not sure how opinionated to be here.

src/db/OrganizationTypes.ts
```diff
@@ -0,0 +1,118 @@
...
+   **/
+  content: IOrganizationContent
+  /**
+   * If this organization has been edited, this field will contain the metadata about the
+   * last edit that was made.
+   */
+  _change?: ChangeRecordMetadataType
+  /** Used to delete an area.  See https://www.mongodb.com/docs/manual/core/index-ttl/ */
+  _deleting?: Date
+  createdAt?: Date
```
CocoisBuggy: Why would this be an optional field? The time of creation is always knowable at the time of creation

zichongkao: Good point! I could enforce that these be mandatory, and have tests pass with minimal fixes. Made the change!

src/graphql/common/MuuidScalar.ts
vnugent, line 1: I've tried adding custom scalar to handle MUUID in the past but ended up removing it because it's not well supported from the frontend

https://github.com/apollographql/apollo-client/issues/8857

I'm not sure if this is specific to Apollo client or across the board?  Can you work with this scalar in Hasura graphql explorer?

src/db/OrganizationTypes.ts
vnugent: I think it's a matter of style. I'm indifferent.

vnugent:
> However, for the public-facing ids, eg area_id, orgId, etc, we use MUUIDs, which are Mongo's primitive UUID construct. As with most UUIDs, these are stored in the backend in binary, but when we handle them in the frontend and make requests with them, they need to be strings. There are two ways to serialize these as strings (which is very confusing) -- one is the "relaxed" mode where they turn into hex strings like "59f1d95a-627d-4b8c-91b9-389c7424cb54" and another is the "canonical" mode where they turn into base64 strings like "WfHZWmJ9S4yRuTicdCTLVA==". It seems we prefer to use "relax" so I've made that consistent everywhere.

You're right. per [uuid-mongo readme](https://github.com/cdimascio/uuid-mongodb#modes):  **Mode only impacts how JSON.stringify(...) represents a UUID**  but we still store it as Mongo's binary.  In the beginning I exposed Mongo ObjectID to the frontend then I thought UUID v4 format would look more recognizable than a random string/number.  Hence the adoption of uuid-mongo library.

> My proposal now is to use MUUIDs across the board except when they are the Mongoose model _id, which shouldn't be publicly accessible and aren't used widely anyway.

Agreed!

src/graphql/common/MuuidScalar.ts
zichongkao: My understanding of how the custom scalar works is that it allows the backend to receive strings eg. `{ input: { orgId: "59f1d95a-627d-4b8c-91b9-389c7424cb54" }}` and it'll convert them to binary. So that inside the mongoose data source, when we access `input.orgId` that will already be a binary object and we don't have to manually convert it.

This means that the client's shouldn't be affected right? They'll send and receive strings and not need to deal with binary or custom scalars. I guess maybe Hasura will throw an error from some sort of MUUID <> string type mismatch. I'll try it out on the `area.metadata.area_id` which is already an MUUID and report back!

PR #255, opened 2023-04-04
feat: Australian grade context

musoke:
This PR defines the Australian grade context.

depends on https://github.com/OpenBeta/sandbag/pull/86

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "@babel/runtime": "^7.17.2",
     "@graphql-tools/schema": "^8.3.1",
-    "@openbeta/sandbag": "^0.0.33",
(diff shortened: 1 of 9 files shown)


musoke:
I'm surprised that these tests passed - @vnugent, do I need to add sample data with AU context somewhere?

vnugent:
Can you please create a test in [MutableClimbDataSource](
https://github.com/OpenBeta/openbeta-graphql/blob/develop/src/model/__tests__/MutableClimbDataSource.ts)?

1.  Create a fictional Sport climbing crag in Australia.  Add new climbs to it.  Make sure grades are parsed correctly (see YSD & French tests)
2.  Create a bouldering area.  Repeat similar tests.

musoke:
Thanks for the pointer - I think this is ready now.

PR #257, opened 2023-04-16
Add _change checks to integration tests

zichongkao:
Addresses: https://github.com/orgs/OpenBeta/projects/6/views/1?pane=issue&itemId=25526852

Confirms that the in-memory mongo replset can support our integration tests. We were worried that the stream listeners required to support change histories would not be supported by the in-memory replset. In this PR, we update the integration tests to demonstrate that the stream listeners do capture incoming changes.

### DIFF
--- a/src/__tests__/integration.ts
+++ b/src/__tests__/integration.ts
@@ -9,8 +9,10 @@ import jwt from 'jsonwebtoken'
 import MutableAreaDataSource, { createInstance as createAreaInstance } from '../model/MutableAreaDataSource.js'
 import MutableOrganizationDataSource, { createInstance as createOrgInstance } from '../model/MutableOrganizationDataSource.js'
 import { AreaType } from '../db/AreaTypes.js'
-import { OrgType, OrganizationType } from '../db/OrganizationTypes.js'
+import { OrgType, OrganizationType, OperationType } from '../db/OrganizationTypes.js'
+import { changelogDataSource } from '../model/ChangeLogDataSource.js'
 
+jest.setTimeout(60000)
 const PORT = 4000
 
 interface QueryAPIProps {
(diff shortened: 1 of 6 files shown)


src/__tests__/integration.ts
```diff
@@ -159,6 +161,27 @@ describe('graphql server', () => {
...
+
+      // history is shown most recent first
+      const updateRecord = orgHistory[0].changes
+      expect(updateRecord[0].dbOp).toEqual('update')
+      expect(updateRecord[0].fullDocument._change?.historyId).toEqual(orgHistory[0]._id) // should point to current change
+      expect(updateRecord[0].fullDocument.displayName).toBe('Allies of Openbeta')
+
+      const createRecord = orgHistory[1].changes
+      expect(createRecord[0].dbOp).toEqual('insert')
+      expect(createRecord[0].fullDocument._change?.historyId).toEqual(orgHistory[1]._id) // should point to current change
```
zichongkao, line 183: In `src/model/__tests__/AreaHistoryDataSource.ts` we check `prevHistoryId`, but in our earlier PR, you mentioned that we're no longer supporting this feature, so we didn't implement this for Organizations therefore there is no `prevHistoryId` to check for. Just wanted to confirm that this understanding is accurate!

src/db/index.ts
```diff
@@ -79,10 +80,10 @@ export const gracefulExit = async (exitCode: number = 0): Promise<void> => {
   })
 }
 
-export const defaultPostConnect = async (): Promise<void> => {
+export const defaultPostConnect = async (): Promise<ChangeStream> => {
   console.log('Kudos!')
   await createIndexes()
-  await streamListener()
+  return await streamListener()
```
zichongkao, line 86: Hope this is ok. Let me know if there's a better design pattern to use to record the stream so we can close it later.

src/__tests__/integration.ts
vnugent: we can remove prevHistoryId

Issue #259, opened 2023-04-16
Ensure that contributors are using the same linter.


bradleyDean:
As we [discussed on the discord server](https://discord.com/channels/815145484003967026/815145898908712980/1096878853588258932), package.json was pointing to the global ts-standard. 
We should consider changing it to `"lint:" "yarn ts-standard"`, so that the correct dependency is used.

PR #260, opened 2023-04-16
Ensure the specified ts-standard dependency is used.

bradleyDean:
This PR solves [this issue](https://github.com/OpenBeta/openbeta-graphql/issues/259).

This change ensures that the version of ts-standard specified in the package.json file will be used, rather than the global version.


vnugent:
Thanks @bradleyDean for the fix.

@all-contributors add @bradleyDean for code & ideas

Issue #262, opened 2023-04-17
Edit Readme to explain commit warnings.


bradleyDean:
New contributors might waste time trying to resolve some warnings when comitting changes, which it turns out we are just ignoring for the time being.
When comitting changes, the following warnings appear:

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.5.0

YOUR TYPESCRIPT VERSION: 4.9.5

Please only submit bug reports when using the officially supported version.

=============
ts-standard: Typescript Standard Style! (https://github.com/standard/ts-standard)
ts-standard: Run `ts-standard --fix` to automatically fix some problems.

```

This issue proposes that we include some explanatory text in the readme that encourages contributors to ignore this warning for now.

If there is already a ticket about fixing the underlying conflict, we should probably link this issue to that one.

musoke:
How hard is it to just fix the potential incompatibility?  This would make on-boarding smoother.

PR #263, opened 2023-04-17
Explain expected warnings generated when commit changes.

bradleyDean:
When committing a change, a warning message shows up in the terminal, as described in [this issue](https://github.com/OpenBeta/openbeta-graphql/issues/262).

This commit implements the proposed changes to the readme, by supplying the explanatory text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,19 @@
 [![Build](https://gitlab.com/openbeta/openbeta-graphql/badges/develop/pipeline.svg)](https://gitlab.com/openbeta/openbeta-graphql/-/pipelines) [![License](https://img.shields.io/github/license/openbeta/openbeta-graphql?style=flat-square)](./LICENSE)
+
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
+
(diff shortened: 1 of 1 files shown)


zichongkao:
Thanks for helping folks understand these errors. But given https://github.com/OpenBeta/openbeta-graphql/pull/265, this isn't necessary anymore right?

bradleyDean:
> Thanks for helping folks understand these errors. But given #265, this isn't necessary anymore right?

Oh, excellent! Yes, I think this solves the problem. I haven't committed any changes yet (soon hopefully), so I cannot personally confirm that the issue is resolved, but it sure looks to me like #265 should do the trick!

musoke:
Yeah, hopefully this PR isn't necessary now, but let's keep it around until @bradleyDean has a chance to verify.

bradleyDean:
> Yeah, hopefully this PR isn't necessary now, but let's keep it around until @bradleyDean has a chance to verify.

It is totally fine with me if someone else verify's. I won't be able to work on this until Saturday, unfortunately.
If anyone else commits a change and doesn't see the warning, I'm sure that we can close this issue.

PR #270, opened 2023-04-25
All fields accessible during addOrganization

zichongkao:
Previously, `addOrganization` would only add a shell object containing `displayName`, and `orgType`. We would have to make a separate `updateOrganization` call later on to add `website`, `instagramLink` and others.

This change enables us to supply `website`, and other fields immediately in the `addOrganization` call, thus reducing one API call.

This PR completes the second outstanding task required for https://github.com/OpenBeta/open-tacos/pull/788

### DIFF
--- a/src/__tests__/integration.ts
+++ b/src/__tests__/integration.ts
@@ -230,28 +230,32 @@ describe('graphql server', () => {
     let charlieOrg: OrganizationType
 
     beforeEach(async () => {
-      let emptyOrg = await organizations.addOrganization(user, 'Alpha Club', OrgType.localClimbingOrganization)
       let document: any = {
+        displayName: 'Alpha Club',
         email: 'admin@alpha.com',
         associatedAreaIds: [ca.metadata.area_id, wa.metadata.area_id]
       }
-      alphaOrg = await organizations.updateOrganization(user, emptyOrg.orgId, document)
+      alphaOrg = await organizations.addOrganization(user, OrgType.localClimbingOrganization, document)
         .then((res: OrganizationType | null) => {
           if (res === null) throw new Error('Failure mocking organization.')
           return res
         })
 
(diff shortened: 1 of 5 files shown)


src/__tests__/integration.ts
```diff
@@ -230,28 +230,32 @@ describe('graphql server', () => {
...
           if (res === null) throw new Error('Failure mocking organization.')
           return res
         })
 
-      emptyOrg = await organizations.addOrganization(user, 'Beta Club', OrgType.localClimbingOrganization)
-      document = { email: 'admin@beta.com' }
-      betaOrg = await organizations.updateOrganization(user, emptyOrg.orgId, document)
+      document = {
+        displayName: 'Beta Club',
+        email: 'admin@beta.com'
```
musoke, line 246: isn't it good practice to avoid using domains we don't own in tests? alpha.com in the above example is  a registered domain.  Unlikely to be a problem since we aren't sending emails now, but might be in future?

musoke:
I'm not familiar with the code being modified here, but all the changes look sensible.  I left one bikeshedy comment - don't consider it a blocker.

musoke (approved):

vnugent:
Looks good to me!

src/__tests__/integration.ts
vnugent: in the future  what about test@example.com or test@openbeta.io?  the later is actually valid.  It goes to a catch-all mailbox.

musoke:
Sounds good to me!

zichongkao:
I fixed the test email addresses in https://github.com/OpenBeta/openbeta-graphql/pull/272 so they point to unused domains just to be extra safe.

I don't really think it's necessary to test using domains we own. After all this is a backend system and the test results don't get rendered anywhere where someone or some automated test could mistakenly access it. This helps readability of tests, and reduces chances of `test@openbeta.io` collisions.

PR #277, opened 2023-04-29
Add more typing for history objects

zichongkao:
- Changes `kind` to be an enum instead of string. 
- Add exhaustive checks at switch statements.
- Add `OrganizationType` as a `SupportedCollectionType`.

### DIFF
--- a/src/db/edit/streamListener.ts
+++ b/src/db/edit/streamListener.ts
@@ -4,10 +4,11 @@ import dot from 'dot-object'
 
 import { changelogDataSource } from '../../model/ChangeLogDataSource.js'
 import { logger } from '../../logger.js'
-import { BaseChangeRecordType, ResumeToken, UpdateDescription, DBOperation, SupportedCollectionTypes } from '../ChangeLogType.js'
+import { BaseChangeRecordType, ResumeToken, UpdateDescription, DBOperation, SupportedCollectionTypes, DocumentKind } from '../ChangeLogType.js'
 import { checkVar } from '../index.js'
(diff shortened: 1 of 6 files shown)


src/graphql/history/HistoryFieldResolvers.ts
```diff
@@ -24,16 +25,16 @@ const resolvers = {
...
-        return 'Area'
+      switch (node.kind) {
+        case DocumentKind.areas:
+          return 'Area'
+        case DocumentKind.climbs:
+          return 'Climb'
+        case DocumentKind.organizations:
+          return 'Organization'
+        default:
+          return exhaustiveCheck(node.kind)
```
zichongkao, line 36: Don't know if there's a better way to do this!

src/db/edit/streamListener.ts
```diff
@@ -49,7 +50,7 @@ const onChange = (change: ChangeStreamDocument): void => {
     case 'replace':
     case 'update': {
       let dbOp: DBOperation = 'update'
-      const source = change.ns.coll
+      const source = DocumentKind[change.ns.coll]
```
zichongkao, line 53: Hope this is fine. Dunno if we can type `change.ns.coll` as well.

vnugent:
Thanks for the fix.  In general I think we should prioritize climb & area history.  As you said in the other comment the LCO edit feature will be used by very a few trusted individual.

vnugent:
@all-contributors add @zichongkao @musoke for code & ideas

PR #283, opened 2023-05-03
feat: Ice context

musoke:
This adds basic support for ice climbs with the WI grade scales.

This currently assumes that all contexts use only Winter Ice grades rather than a mix of Alpine Ice and Winter Ice.  This is not the case; the grade scale depends on the local climate.

To enable AI grades in future, the ice grades allowed in most contexts should be an array: `[GradeScales.AI, GradeScales.WI]`.

----

Depends on 

- [x] bump version of sandbag
- [x] update version requirement here

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@babel/runtime": "^7.17.2",
     "@google-cloud/storage": "^6.9.5",
     "@graphql-tools/schema": "^8.3.1",
-    "@openbeta/sandbag": "^0.0.37",
+    "@openbeta/sandbag": "^0.0.45",
     "@turf/area": "^6.5.0",
     "@turf/bbox": "^6.5.0",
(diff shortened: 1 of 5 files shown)


vnugent:
@musoke @zichongkao can we revisit this PR?

musoke:
Yup.  I was waiting for things to settle with the LCO org and picture rewrite.

musoke:
@actuallyyun, could you take a look at the errors here?  I'm updating the version of sandbag here and getting errors relating to `GradeScales` exports.  I'm wondering if this is related to your changes in https://github.com/OpenBeta/sandbag/pull/106

musoke:
~blocked by https://github.com/OpenBeta/sandbag/issues/123~

vnugent (approved):

Issue #267, opened 2023-04-19
Implement developer authentication bypass for local development


bradleyDean:
**Note:** This description has been edited. It originally (accidentally) blended front end and backend authentication into a single issue. I will link to the front end authentication bypass issue here, as soon as I write up that issue.

**The Problem:**
As developers working on the openbeta-graphql project, we need a convenient way to bypass the authentication system when running the server locally. This will allow us to test/develop functionality without having to provide real user credentials. Note that [this issue](https://github.com/OpenBeta/open-tacos/issues/798) describes creating test/dev accounts which could probably be used to authenticate backend request for dev purposes, but I think that process would be clunky and it is probably best not to rely on that strategy.

**Proposed Solution Overview:**
Allow contributors to bypass the authentication system when running the app locally. We'll achieve this by setting a new, `NODE_ENV` environment variable to `"development"` when running a new `serve-dev` command, which will run the app in development mode.

**Implementation Details:**

1. Add a new script entry (below [this line](https://github.com/OpenBeta/openbeta-graphql/blob/7375cc509c0c71135a12d10ee30a942fba333510/package.json#L65)) called `serve-dev` in package.json:

```
"scripts": {
  ...
  "serve": "yarn build && node --experimental-json-modules build/main.js",
  "serve-dev": "NODE_ENV=development yarn serve",
  ...
}
```

2. Update the [createContext function](https://github.com/OpenBeta/openbeta-graphql/blob/7375cc509c0c71135a12d10ee30a942fba333510/src/auth/middleware.ts#L16-L26) in `build/auth/middleware.js` to bypass the authentication process when `process.env.NODE_ENV` is set to development. Assign 'admin' and 'editor' roles to the user in this case.
```
// ...(existing logic)
  if (process.env.NODE_ENV === 'development') {
    user.roles = ['admin', 'editor'];
    user.uuid = 'some-test-uuid';
  } else {
    const authHeader = String(headers?.authorization ?? '');
    if (authHeader.startsWith('Bearer ')) {
      // ... (existing logic)
```
3. (Optional) Write tests to cover the updated `createContext` function.

**Acceptance Criteria**:

- A separate script called `serve-dev` should be available for running the app in development mode.
- The app should allow developers to bypass the authentication system when running locally with the serve-dev script.
- Developers should be automatically assigned the roles of 'admin' and 'editor' in the development environment when using the serve-dev script.
- (Optional )The updated `createContext` function should be covered by tests.

[...]

PR #286, opened 2023-05-05
Bradley/dev mode credentials

bradleyDean:
**What?**
Create `serve-dev` command which automatically authenticates the user and gives them certain permissions, as described in [this issue](https://github.com/OpenBeta/openbeta-graphql/issues/267).

**How?**
The issue linked above described the process pretty well, but notice that `createContext` now creates a closure so that seperate requests will not keep creating new user id's. I also created a launch config object so we can run the new `serve-dev` command using a debugger.

**Validation?**
I used the `serve-dev` command to add areas to the db in the graphql playground. I was shaky on how to get the necessary uuid's (for specifying which specific area I wanted to delte, fetch, etc), so it would probably be a good idea for someone more familiar with the API to run the new command, then exercise other API endpoints, just so we can validate it with more code paths. Alternatively, maybe someone more familiar with using the API and I can chat on discord and I'll run additional validation myself.

**Tests?**
I haven't implemented any tests yet :(
If anyone thinks I should, let me know and I'll get on it.

### DIFF
--- a/src/auth/middleware.ts
+++ b/src/auth/middleware.ts
@@ -23,7 +23,7 @@ export const createContext = (() => {
     }
 
     if (process.env.GOD_MODE === 'true' && (user.uuid == null)) {
-      user.roles = ['user_admin','org_admin', 'editor']
+      user.roles = ['user_admin', 'org_admin', 'editor']
       user.uuid = testUUID
     } else {
       const authHeader = String(headers?.authorization ?? '')
--- a/src/auth/middleware.ts
+++ b/src/auth/middleware.ts
@@ -1,4 +1,5 @@
 import muid, { MUUID } from 'uuid-mongodb'
+import { logger } from '../logger.js'
 import { AuthUserType } from '../types.js'
 import { verifyJWT } from './util.js'
 
@@ -8,9 +9,9 @@ import { verifyJWT } from './util.js'
 export const createContext = (() => {
   let testUUID: MUUID
 
-  if (process.env.NODE_ENV === 'development') {
+  if (process.env.GOD_MODE === 'true') {
     testUUID = muid.v4()
-    console.log(`The user.uuid for this session is: ${testUUID.toString()}`)
+    logger.info(`The user.uuid for this session is: ${testUUID.toString()}`)
   }
 
   return async ({ req }): Promise<any> => {
@@ -21,8 +22,8 @@ export const createContext = (() => {
       uuid: undefined
     }
 
-    if (process.env.NODE_ENV === 'development' && (user.uuid == null)) {
-      user.roles = ['admin', 'editor']
+    if (process.env.GOD_MODE === 'true' && (user.uuid == null)) {
+      user.roles = ['user_admin','org_admin', 'editor']
       user.uuid = testUUID
     } else {
       const authHeader = String(headers?.authorization ?? '')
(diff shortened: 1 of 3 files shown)


bradleyDean:
I converted to a draft because I decided to go ahead and write some tests for this. 
I do not plan on changing the logic in the diffs (unless the tests reveal bugs!), so it shouldn't be a waste of time to have a look at the diffs in the mean time!

Update:
I ran into issues writing tests, having to do with a tricky race condition with the setting of the `env.GOD_MODE` variable. I am moving forward without resolving this for now, but if anyone is concerned about it, let me know and I'll put more time into figuring it out.

I did manually test that I can use `serve-dev` to autenticate mutations involving areas and orgs.

src/auth/middleware.ts
```diff
@@ -1,27 +1,41 @@
...
-  const { headers } = req
+export const createContext = (() => {
+  let testUUID: MUUID
 
-  const user: AuthUserType = {
-    roles: [],
-    uuid: undefined
+  if (process.env.NODE_ENV === 'development') {
+    testUUID = muid.v4()
+    console.log(`The user.uuid for this session is: ${testUUID.toString()}`)
```
vnugent: Can you please use `logger.info()` instead?

src/auth/middleware.ts
```diff
@@ -1,27 +1,41 @@
...
-    user.roles = z?.['https://tacos.openbeta.io/roles'] ?? []
-    const uidStr: string | undefined = z?.['https://tacos.openbeta.io/uuid']
-    user.uuid = uidStr != null ? muid.from(uidStr) : undefined
-  }
+    const user: AuthUserType = {
+      roles: [],
+      uuid: undefined
+    }
+
+    if (process.env.NODE_ENV === 'development' && (user.uuid == null)) {
```
vnugent: What do you think if we require the user to explicitly opt-in instead of relying on the default `NODE_ENV` var?  Maybe create a new variable "GOD_MODE=true" or LIZARD_MODE or whatever you want to call it, and set it before calling `dev-serve` in package.json. 

My concern is in production we might inadvertently unset `NODE_ENV=production` or forget to set it when working with the docker file or devops automation task.

vnugent (changes requested):

src/auth/middleware.ts
```diff
@@ -1,27 +1,41 @@
...
+      user.roles = ['admin', 'editor']
```
zichongkao: FYI the three roles that currently exist are `editor`, `user_admin` and `org_admin`.

src/auth/middleware.ts
bradleyDean: I think this is a good idea. I became concerned about this myself when I noticed that while the tests are running, NODE_ENV is set to 'test'. I'll change it to GOD_MODE=true.

I am not sure what you mean by "before calling `dev-serve`". I'm going to stick with the current strategy, where `serve-dev` sets the env variable, but if you want me to do something else, let me know and I'll change it.

src/auth/middleware.ts
bradleyDean: Ok, I changed it to include all three roles.

src/auth/middleware.ts
vnugent: > I am not sure what you mean by "before calling dev-serve". I'm going to stick with the current strategy, where serve-dev sets the env variable, but if you want me to do something else, let me know and I'll change it.

Sorry I mixed up serve-dev and dev-serve.  This is what I meant:

package.json
```json
"serve-dev": "GOD_MODE=true yarn serve"
```

vnugent:
@bradleyDean can you please rebase to fix conflicts?

l4u532:
Should be closed since this got implemented by #371

Issue #202, opened 2022-12-06
Resolving User fields within the GQL schema


CocoisBuggy:
## Why 🤔

The ability to have a comprehensive entity schema is the **main** upside of GraphQL. Being able to compose queries of all kinds is a massive bonus, and will encourage developers to implement more features on [OpenTacos](https://github.com/OpenBeta/open-tacos) with a vastly softer learning curve - and a much easier local development setup.

leaving these resolves to Next `/api` calls is an under-utilization of GraphQL's core offering that may result in over lock-in to the NextJS framework if not addressed now.

## Problem Spec

As the offering for Posts and Contributions evolve, having a more comprehensive GQL schema definition will improve developer ergonomics and could potentially reduce developer on-boarding friction in the future.

```gql
type History {
    id: ID!
    editedBy: String!
    operation: String!
    createdAt: Date!
    changes: [Change]
 }
```
Gives sufficiently authorized developers a road to resolving users themselves, but adds an unreachable or frustrating step for new devs who are unfamiliar with the service structure.

`editedBy: String` also does not communicate anything about the schema data relationship - another benefit of GQL schemas 

### Current model

I believe the current model would have NextJS do an additional fetch/join to get this data from Auth0. The proposal made here is to move that logic into the resolver to improve API ergonomics for current and incoming API functionality. The upcoming post mechanics being an example. 

Resolving within a single GQL endpoint is going to become increasingly desirable as contributions become more fleshed out, and we will inevitably end up with masses of relational data in the mongo data-store (Ticks, Posts, media, change-history, comments). 

_Additionally, if we were ever to experiment with alternate U.I frameworks we would need to commit to much more work and technical risk, than if principal data were all available through a single GQL endpoint._

## Proposed Solution

Add an additional type to the schema, making a more holistic experience for API consumers.

The type would naturally not expose any private user data, and would return the same data available within the various api handlers already extant within OpenTacos.

```gql
type User {
  "Globally unique static ID for this user"
  id: ID!
  "Globally unique, but mutable"
  userName: String!
  "If an Avatar exists for this user, it is available here"
  avatarUrl: String
}
```

which could allow resolution for current and future types in the schema:

```gql
type History {
    id: ID!
    editedBy: User!
    operation: String!
    createdAt: Date!
    changes: [Change]
 }

# example of another type that could use user, but does not exist
type Media {
   id: ID!
   author: User!
   tags: [MediaTagType!]
}
```

[...]

PR #290, opened 2023-05-11
fi: support uuid -> username mapping

vnugent:
Issue #202 

Incrementally move user profile data to our own db.  The goal of this PR is to support syncing username from Auth0 to our db to enable mapping from user uuid to username.

- [x] support user profile create/update
- [x] support uuid -> username mapping
- [x] restrict setting email field on profile creation
- [x] introduce "builder service account" with elevated permissions to read user profile
- [x] retire get username hack
- [x] fix unit tests
- [ ] run job to create users collection in prod


vnugent:
@zichongkao any idea why end-to-end tests are failing?

zichongkao:
> @zichongkao any idea why end-to-end tests are failing?

Basically in https://github.com/OpenBeta/openbeta-graphql/blob/2589e6be07c2208535480dee70177477238ae03a/src/utils/testUtils.ts#L26 we mock `jwt.verify`'s return value and the return value has no `scope` field. I made the middleware robust to not having `scope`. But in future, I'm sure we'll have to mock `scope` in the mocked response so that we can give `isBuilder` status for testing out user profile stuff.

zichongkao:
I'll review the PR later this evening!

vnugent:
I pulled your fix down.  API tests are still failing inside `queryAPI()` 

```
  ● areas API › queries › retrieves an area and lists associated organizations

    expect(received).toBe(expected) // Object.is equality

    Expected: 200
    Received: 500

      80 |         userUuid
      81 |       })
    > 82 |       expect(response.statusCode).toBe(200)
         |                                   ^
```

The response object:

```
{
        status: 500,
        text: '{"errors":[{"message":"Context creation failed: connect ECONNREFUSED 127.0.0.1:80","extensions":{"code":"INTERNAL_SERVER_ERROR"}}]}\n',
        method: 'POST',
        path: '/'
      }

      at Object.<anonymous> (src/__tests__/areas.ts:83:15)

  ● are

```

zichongkao:
That's a bit surprising. They pass locally and on the automated checks. Is your .env in a good state? On my end, I don't see anything that is trying to connect to port 80.

vnugent:
> That's a bit surprising. They pass locally and on the automated checks. Is your .env in a good state? On my end, I don't see anything that is trying to connect to port 80.

You're right.  I have a `.env.local` that overrides some of the DB values.  All green now.

src/__tests__/areas.ts
```diff
@@ -79,10 +79,11 @@ describe('areas API', () => {
         variables: { input: wa.metadata.area_id },
         userUuid
       })
+
       expect(response.statusCode).toBe(200)
       const areaResult = response.body.data.area
       expect(areaResult.uuid).toBe(muuidToString(wa.metadata.area_id))
-      expect(areaResult.organizations.length).toBe(1)
+      expect(areaResult.organizations).toHaveLength(1)
```
zichongkao, line 86: Thanks for fixing this!

src/db/UserSchema.ts
```diff
@@ -18,6 +18,53 @@ export const ExperimentalUserSchema = new Schema<ExperimentalUserType>({
...
+})
+
+export const UserSchema = new Schema<User>({
+  _id: {
+    type: 'object',
+    value: { type: 'Buffer' }
+  },
+  email: { type: Schema.Types.String },
+  emailVerified: { type: Schema.Types.Boolean },
+  displayName: { type: Schema.Types.String },
```
zichongkao, line 46: What's the difference between `displayName` and `username`? And why do we track timestamps on `username` specifically? Shouldn't the UserSchema's default timestamps enabled by `timestamps: true` track these already?

src/db/UserSchema.ts
```diff
@@ -18,6 +18,53 @@ export const ExperimentalUserSchema = new Schema<ExperimentalUserType>({
...
 export const getExperimentalUserModel = (): mongoose.Model<ExperimentalUserType> => {
   return mongoose.model('exp_users', ExperimentalUserSchema)
 }
+
+const UsernameSchema = new Schema<UsernameInfo>({
+  username: { type: Schema.Types.String }
+}, {
+  _id: false,
+  timestamps: {
+    updatedAt: true,
```
zichongkao, line 34: Slightly confused too, since below we type this as a "NotUpdatableField", so how come we allow an `updatedAt` time?

src/db/UserSchema.ts
```diff
@@ -18,6 +18,53 @@ export const ExperimentalUserSchema = new Schema<ExperimentalUserType>({
   timestamps: true
 })
 
+/**
+ * Temporary model used to capture user profile during bulk import of Canada data.
+ * Use the standard User model instead.
+ */
 export const getExperimentalUserModel = (): mongoose.Model<ExperimentalUserType> => {
```
zichongkao, line 25: Just out of curiosity -- will this go away after the migration of more data to our new User collection is complete?

src/model/UserDataSource.ts
```diff
@@ -0,0 +1,200 @@
...
+      email,
+      emailVerified,
+      _id
+    } = input
+
+    if (Object.keys(input).length === 0) {
+      throw new Error('Nothing to update. Must provide at least one field.')
+    }
+
+    const username = trimToNull(_username)
```
zichongkao, line 47: It might be a bit confusing to the user if they try to add a username like "  doublespaceman  " which would be trimmed to "doublespaceman" and saved. The user would end up with a username they don't expect. It might reduce confusion to validate without trimming and return an error straight away.

src/model/UserDataSource.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    const rs = await this.userModel.findOne({ _id })
+
+    const isNew = rs == null
+
+    if (isNew) {
+      if (email == null) {
+        throw new Error('Email is required when creating a new profile')
+      }
+      let usernameInfo: UsernameInfo | null = null
```
zichongkao, line 73: I'm also wondering -- username seems so central. Why don't we want to force creation of a username right off the bat? Furthermore, we could force a user-inputted rather than auto-generated username to avoid issues like this: https://github.com/OpenBeta/open-tacos/issues/750. A bit more friction, but probably makes the site seem more human when there are human names everywhere as opposed to robot-like names.

src/db/UserSchema.ts
zichongkao: Oh i see, it's to facilitate a dedicated cooldown time for username changes right?  I guess the alternative would be to parse the changeLog, but I can see why that would be a pain in the ass.

And OK - I see now that `username` is basically like the insta handle @zckao say, and `displayName` is what we show on the profile "Zi Chong Kao". Perhaps "handle" or "userHandle" might disambiguate the two better. But don't feel strongly, and would require changing many many references in the codebase -- so probably not worth it.

src/model/UserDataSource.ts
```diff
@@ -0,0 +1,200 @@
...
+export default class UserDataSource extends MongoDataSource<User> {
+  userModel = getUserModel()
+
+  /**
+   * Update user profile.  Create a new user object if not defined.
+   * @param userUuid userUuid to update/add
+   * @param input profile params
+   * @returns true if successful
+   */
+  async createOrUpdateUserProfile (updater: MUUID, input: UpdateProfileGQLInput): Promise<boolean> {
```
zichongkao: Is the idea that this would be called at signup time, alongside the current call to Auth0? And is it accurate that the long-term vision is to slowly move more and more data that is stored in the Auth0 object over into this model?

zichongkao (commented):
Looks good in general! I can see how adding username resolvers will help reduce frontend API calls in many places.

Many of the questions I raised are for my own learning.

The one product decision that might be worth discussing is whether to make usernames mandatory during sign-up. I'm leaning toward yes, especially since they are mutable so users can revisit them later if they don't like them. Otherwise the power of defaults is strong and users may just leave the auto-generated ones.

src/db/UserSchema.ts
zichongkao: Ok I understand now why we allow `updatedAt`. But still wondering why we list as a "NotUpdatableField".

vnugent: `updatedAt: true`: Mongo will set the value for you whenever there's an update
*Edit: in this case I simply want Mongo to create the updatedAt field for me.  We want to set username update timestamp ourselves, but don't want API users to do it.

src/model/UserDataSource.ts
vnugent: Correct.  I'll remove the auto-generated username in Auth0 login action.  In the new flow, new users will be directed to a new username creation screen (frontend issue https://github.com/OpenBeta/open-tacos/issues/830). New users can't skip/bypass this screen because if a user is logged in, the frontend will always call `getUsername()` to see if the usernameInfo object has set.

src/db/UserSchema.ts
vnugent: I created this 'experimental' collection for a volunteer to bulk add Canada data. The intent was for preserving MP author information.  Yes, we can look into migrating the data over at some point.

src/model/UserDataSource.ts
vnugent: As discussed in the other comment, new open-tacos users will be required to create a username.  Allowing usernameInfo object to be nullable in the backend will let us create new user accounts programmatically (a future use case).

src/model/UserDataSource.ts
vnugent: moved username validation above trimming

asythlos:
cutegirl113651 hack this insta account

Issue #264, opened 2023-04-18
Improve Types: Specify dataSources return value type.


bradleyDean:
[`dataSources`](https://github.com/OpenBeta/openbeta-graphql/blob/7375cc509c0c71135a12d10ee30a942fba333510/src/server.ts#L25) is a function with type: `() => Datasources<object> | undefined`.

I propose that we specify a type for that generic parameter, say, `DataSourcesValue` or something like that, so the type becomes:
  `() => Datasources<DataSourceValue> | undefined`.
  
This way, we can have a helpful type definition for other references to that value. Specifically, I was looking [here](https://github.com/OpenBeta/openbeta-graphql/blob/7375cc509c0c71135a12d10ee30a942fba333510/src/graphql/resolvers.ts#L75), where it seems that the parameter, `dataSource`, now refers to an object, not a function. I am pretty sure this parameter refers to an object with the type that I called `DataSourceValue`, above. 

Maybe while we're at it, we could add a type for [`context`](https://github.com/OpenBeta/openbeta-graphql/blob/7375cc509c0c71135a12d10ee30a942fba333510/src/graphql/resolvers.ts#L96), too. I think such types are helpful because as a new contributor, I search the code base for types to help me figure out where the arguments to these sorts of functions are coming from. 

What do y'all think?

PR #291, opened 2023-05-11
Use Apollo Server Context type more widely

zichongkao:
In response to https://github.com/OpenBeta/openbeta-graphql/issues/264

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -12,28 +12,31 @@ import { createContext, permissions } from './auth/index.js'
 import XMediaDataSource from './model/XMediaDataSource.js'
 import PostDataSource from './model/PostDataSource.js'
 import { createInstance as createNewOrgDS } from './model/MutableOrganizationDataSource.js'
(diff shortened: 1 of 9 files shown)


zichongkao:
@bradleyDean What do you think?

src/server.ts
```diff
@@ -12,28 +12,31 @@ import { createContext, permissions } from './auth/index.js'
...
 import { createInstance as createNewOrgDS } from './model/MutableOrganizationDataSource.js'
+import type { Context } from './types.js'
+import type { DataSources } from 'apollo-server-core/dist/graphqlOptions'
 
 export async function createServer (): Promise<ApolloServer> {
   const schema = applyMiddleware(
     graphqlSchema,
     permissions.generate(graphqlSchema)
   )
+  const dataSources: () => DataSources<Context> = () => ({
```
zichongkao, line 23: This is the key change which specifies the type instead of leaving it as `DataSources<object>`.

PR #292, opened 2023-05-13
Add organizations field in area resolver

zichongkao:
Enables us to retrieve organizations associated with an area in a single area query. Eg:
```
query area($input: ID) {
  area(uuid: $input) {
    uuid
    organizations {
      ...
    }
  }
}
```

Under the hood it does two things in the area resolver:
1. Finds organizations with an `associatedAreaId` that is the area requested or one of its parents. (I.e. Associating with an area automatically associates with all of its subareas.
2. Remove organizations with an `excludedAreaId` that is the area requested. (I.e. unlike associating, excluding does not exclude the entire subtree. Only the exact area listed.)

### DIFF
--- a/src/__tests__/areas.ts
+++ b/src/__tests__/areas.ts
@@ -1,0 +1,104 @@
+import { ApolloServer } from 'apollo-server'
+import muuid from 'uuid-mongodb'
+import { jest } from '@jest/globals'
+import MutableAreaDataSource, { createInstance as createAreaInstance } from '../model/MutableAreaDataSource.js'
+import MutableOrganizationDataSource, { createInstance as createOrgInstance } from '../model/MutableOrganizationDataSource.js'
+import { AreaType } from '../db/AreaTypes.js'
+import { OrgType, OrganizationType, OrganizationEditableFieldsType } from '../db/OrganizationTypes.js'
+import { queryAPI, setUpServer } from '../utils/testUtils.js'
+import { muuidToString } from '../utils/helpers.js'
+
+jest.setTimeout(60000)
+
+describe('areas API', () => {
+  let server: ApolloServer
+  let user: muuid.MUUID
+  let userUuid: string
+  let inMemoryDB
+
+  // Mongoose models for mocking pre-existing state.
+  let areas: MutableAreaDataSource
+  let organizations: MutableOrganizationDataSource
+  let usa: AreaType
+  let ca: AreaType
+  let wa: AreaType
+
+  beforeAll(async () => {
+    ({ server, inMemoryDB } = await setUpServer())
+    // Auth0 serializes uuids in "relaxed" mode, resulting in this hex string format
+    // "59f1d95a-627d-4b8c-91b9-389c7424cb54" instead of base64 "WfHZWmJ9S4yRuTicdCTLVA==".
+    user = muuid.mode('relaxed').v4()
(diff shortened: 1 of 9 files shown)


src/__tests__/areas.ts
```diff
@@ -0,0 +1,104 @@
...
+    await server.stop()
+    await inMemoryDB.close()
+  })
+
+  describe('queries', () => {
+    const areaQuery = `
+      query area($input: ID) {
+        area(uuid: $input) {
+          uuid
+          organizations {
```
zichongkao, line 53: This is the new query we can now write.

src/__tests__/areas.ts
```diff
@@ -0,0 +1,104 @@
+import { ApolloServer } from 'apollo-server'
```
zichongkao, line 1: New integration tests for areas.

src/graphql/resolvers.ts
```diff
@@ -251,7 +251,20 @@ const resolvers = {
...
       return await media.findMediaByAreaId(node.metadata.area_id, node.ancestors)
     },
-    authorMetadata: getAuthorMetadataFromBaseNode
+
+    authorMetadata: getAuthorMetadataFromBaseNode,
+
+    organizations: async (node: AreaType, args: any, { dataSources }: Context) => {
+      const { organizations } = dataSources
+      const areaIdsToSearch = [node.metadata.area_id, ...node.ancestors.split(',').map(s => muid.from(s))]
+      const associatedOrgs = await (await organizations.findOrganizationsByFilter({
```
vnugent: In [`findOrganizationsByFilter()`](https://github.com/OpenBeta/openbeta-graphql/blob/17ba6ad6dabbc6e5b5e9cb725768cc81085f7f83/src/model/OrganizationDataSource.ts#LL43C33-L43C44) maybe add `toArray()` there and await to avoid 2 awaits here?

vnugent (changes requested):
I'd use `if (filters != null)` to test for nullness and undefined (1 equal sign).  The test would pass here if filters is `null`.

https://github.com/OpenBeta/openbeta-graphql/blob/17ba6ad6dabbc6e5b5e9cb725768cc81085f7f83/src/model/OrganizationDataSource.ts#L16

src/graphql/resolvers.ts
zichongkao: Thanks! Good suggestion. A lot cleaner that way :)

zichongkao: Hmm unfortunately there's another reference that requires the endpoint to return a cursor instead of array:
image

I tried to clean it up and split out the awaits instead. Hopefully a bit more readable. Lmk what you think.

src/graphql/organization/OrganizationQueries.ts
```diff
@@ -23,7 +23,7 @@ const OrganizationQueries = {
     if (sort != null) {
       return await filtered.collation({ locale: 'en' }).sort(sort).limit(limit).toArray()
     } else {
-      return await filtered.collation({ locale: 'en' }).limit(limit).toArray()
+      return await filtered.limit(limit).toArray()
```
zichongkao, line 26: Removed this because it seems collations are there to support sorting of strings.

vnugent:
LGTM!

PR #293, opened 2023-05-18
updateSortingOrder for areas allows re-sorting in-place

zichongkao:
In response to problem noted here: https://github.com/OpenBeta/open-tacos/pull/696#issuecomment-1552148382

> In testing, I'm hitting the duplicate index error for leftRightIndex. I think it's because I'm sending [{areaId: .., leftRightIndex: 0}, {areaId: .., leftRightIndex: 1}, .., {areaId: .., leftRightIndex: <# of childareas -1>}] and when we try to set the leftRightIndex to be 0 for the first item, there is already another entry in the db with leftRightIndex 0, so throws the error even though later on, that item's leftRightIndex will replaced by a different entry. I could reset everything to -1 and then reassign, but it's not that elegant.

In the end, I couldn't find a better way than resetting everything -1. I looked into seeing if Mongo supports deferred constraints to get it to check the uniqueness of `leftRightIndex` only after all the bulk updates are complete. I also looked into whether Mongo could update multiple documents atomically, to avoid the intermediate state where there is a duplicate index. Both didn't work out.

### DIFF
--- a/src/model/MutableAreaDataSource.ts
+++ b/src/model/MutableAreaDataSource.ts
@@ -368,9 +368,25 @@ export default class MutableAreaDataSource extends AreaDataSource {
     const doUpdate = async (session: ClientSession, user: MUUID, input: UpdateSortingOrderType[]): Promise<string[]> => {
       const opType = OperationType.orderAreas
       const change = await changelogDataSource.create(session, user, opType)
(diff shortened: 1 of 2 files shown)


vnugent:
Nice workaround!

Issue #294, opened 2023-05-20
Sort area children by leftRightIndex by default


zichongkao:
This would enable us not to sort on the frontend. See https://github.com/OpenBeta/open-tacos/pull/696#discussion_r1199566295 for more details.

PR #296, opened 2023-05-21
Sort area children by leftRightIndex

zichongkao:
"Area" graphql objects have a children field containing other areas or climbs. Currently these are unsorted -- we leave the frontend to sort them. But the frontend tends to sort them the same way based on their `leftRightIndex` so why not just return that as a default order returned by the backend? This way frontends won't have to do any sorting.

In response to issue: https://github.com/OpenBeta/openbeta-graphql/issues/294 and first discussed in this PR: https://github.com/OpenBeta/open-tacos/pull/696#discussion_r1199566295

### DIFF
--- a/src/graphql/resolvers.ts
+++ b/src/graphql/resolvers.ts
@@ -20,7 +20,7 @@ import { OrganizationMutations, OrganizationQueries } from './organization/index
 import { TickMutations, TickQueries } from './tick/index.js'
 import { UserQueries, UserMutations, UserResolvers } from './user/index.js'
 import { getAuthorMetadataFromBaseNode } from '../db/utils/index.js'
-import { geojsonPointToLatitude, geojsonPointToLongitude } from '../utils/helpers.js'
+import { compareAreaLeftRightIndex, compareClimbLeftRightIndex, geojsonPointToLatitude, geojsonPointToLongitude } from '../utils/helpers.js'
 
 /**
  * It takes a file name as an argument, reads the file, and returns a GraphQL DocumentNode.
@@ -202,7 +202,7 @@ const resolvers = {
 
     children: async (parent: AreaType, _, { dataSources: { areas } }: Context) => {
       if (parent.children.length > 0) {
-        return await areas.findManyByIds(parent.children)
+        return (await areas.findManyByIds(parent.children)).sort(compareAreaLeftRightIndex)
       }
       return []
     },
@@ -221,11 +221,11 @@ const resolvers = {
       // Test to see if we have actual climb object returned from findOneAreaByUUID()
       const isClimbTypeArray = (x: any[]): x is ClimbType[] => x[0].name != null
       if (isClimbTypeArray(result)) {
-        return result
+        return result.sort(compareClimbLeftRightIndex)
       }
 
       // List of IDs, we need to convert them into actual climbs
-      return await areas.findManyClimbsByUuids(result as MUUID[])
+      return (await areas.findManyClimbsByUuids(result as MUUID[])).sort(compareClimbLeftRightIndex)
     },
 
     metadata: (node: AreaType) => {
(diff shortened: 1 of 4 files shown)


src/utils/helpers.ts
```diff
@@ -21,3 +23,17 @@ export function exhaustiveCheck (_value: never): never {
 
 export const geojsonPointToLongitude = (point: Point): number => point.coordinates[0]
 export const geojsonPointToLatitude = (point: Point): number => point.coordinates[1]
+
+export function compareAreaLeftRightIndex (a: AreaType, b: AreaType): number {
```
zichongkao: I considered merging these two, but then I'd need to be able to narrow the type of `a` and `b`. And that's not very easy since these are custom types, not just strings or numbers. And so it gets a bit messy. I figure just splitting them out is a bit repetitive but a lot more readable.

src/graphql/resolvers.ts
```diff
@@ -202,7 +202,7 @@ const resolvers = {
 
     children: async (parent: AreaType, _, { dataSources: { areas } }: Context) => {
       if (parent.children.length > 0) {
-        return await areas.findManyByIds(parent.children)
+        return (await areas.findManyByIds(parent.children)).sort(compareAreaLeftRightIndex)
```
vnugent: What do you think if we let Mongo handle the sorting here and in `findManyClimbsByUuids()`.  This way we don't have to write comparators. 
Something like:

```javascript

// inside AreaDataSource
 ...
 return await this.areaModel.find().
   where('metadata.leftRightIndex').
   in(parent.children).
   sort({ 'metadata.leftRightIndex': 1 }).lean()
```

edit:  add lean()

vnugent (changes requested):

zichongkao:
Yah. That's way better. Db should handle it indeed. I can look at it later
today!

src/model/AreaDataSource.ts
```diff
@@ -108,6 +108,12 @@ export default class AreaDataSource extends MongoDataSource<AreaType> {
           $set: {
             'climbs.gradeContext': '$gradeContext' // manually set area's grade context to climb
           }
+        },
+        {
+          $set: {
+            climbs: { $sortArray: { input: '$climbs', sortBy: { 'metadata.left_right_index': 1 } } },
```
zichongkao, line 123: `$sortArray` requires Mongo 5.2 and up. Otherwise we need something like this: https://stackoverflow.com/questions/15388127/mongodb-sort-inner-array

src/model/AreaDataSource.ts
```diff
@@ -108,6 +108,12 @@ export default class AreaDataSource extends MongoDataSource<AreaType> {
...
+            children: { $sortArray: { input: '$children', sortBy: { 'metadata.leftRightIndex': 1 } } }
```
vnugent, line 124: can you confirm `children[]` contains actual area objects?  I think it only contains a list of area _id's.  When we ask for an area.children[] in GQL, we're making another DB query here:
https://github.com/OpenBeta/openbeta-graphql/blob/develop/src/graphql/resolvers.ts#L203

zichongkao: Oh yeah you're right! I've not tested this far because it throws a "No such operation as $sortArray" error.

zichongkao: What if we did a $lookup for children above this `$set` so that we have the full object?

vnugent: > What if we did a $lookup for children above this $set so that we have the full object?

I thought the query would get too wildly complicated for a small performance gain, whereas the GQL field resolution way [in my comment above](https://github.com/OpenBeta/openbeta-graphql/pull/296#discussion_r1199792692) is more elegant and simpler (but requires an additional query).

vnugent:
I'd be content with whatever solution you come with that doesn't require a DB upgrade, or we can leave this PR open until we do the upgrade.

zichongkao:
Agree that we can hold this back until we have time to upgrade Mongo to v6 as tracked here: https://github.com/OpenBeta/openbeta-graphql/issues/300

Issue #175, opened 2022-10-21
Use Date/timestamp for ticks


vnugent:
Tick date is currently stored as `YYYY-MM-DD` string.  We should store it in the db as Date or seconds from epoch to make it easier to  visualize/sort/filter

 https://github.com/OpenBeta/openbeta-graphql/blob/staging/src/db/TickTypes.ts#L8

CocoisBuggy:
Hey Viet, I was thinking of integrating this with another PR I'm working on but there may be a bit of nuance:

When ticks are added to the database they are not of a guaranteed origin or shape. Most of the unstructured fields in the tick object come from trying to support both MP and OB native ticks. 

I have not looked too deeply into how the tick import system was made, but if the date climbed is a free text field on MP then it is not as simple as converting the field.

Someone will need to verify that MP ticks are structured in this way first, or we will need to perhaps support both date fields

PR #301, opened 2023-05-25
Convert Tick dateClimber from string to date

zichongkao:
https://github.com/OpenBeta/openbeta-graphql/issues/175

I checked that MP ticks are all recorded in 'YYYY-MM-DD' format:
image

And `$dateFromString` (https://www.mongodb.com/docs/v6.0/reference/operator/aggregation/dateFromString/#mongodb-expression-exp.-dateFromString) will be able to convert these strings into Mongo dates.

Main obstacle is writing a test for the migration. This required us to insert data with `dateClimbed` as strings, which is incompatible with the Mongoose schema, so I had to insert it directly into the MongoDB using the Node driver.


src/db/utils/jobs/migration/ConvertTickDateClimbedFromStringToDate.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Converts dateClimbed field on ticks from string to date.
+ */
+const onConnected = async (): Promise<void> => {
+  logger.info('Migrating ticks...')
+  const tickModel = getTickModel()
+
+  await (await tickModel.updateMany(
+    { dateClimbed: { $exists: true } },
+    { $set: { dateClimbed: { $toDate: '$dateClimbed' } } }
```
zichongkao, line 14: Timezone here is a bit tricky. Naive version is to use UTC. But most ticks are probably for climbs in the US, so a first approximation may be to use some US timezone. And then the most labor intensive way to is to discern the timezone where the climb is and adjust accordingly.

zichongkao:
I'm proposing the following deployment sequence:
1. Disabled frontend tick functionality to prevent new data from coming in during the migration and also have input paths start writing Dates not strings: https://github.com/OpenBeta/open-tacos/pull/841
2. We deploy this PR and run the migration.
3. Re-enable frontend tick functionality, and change the output paths to expect `dateClimbed` as Dates not strings: https://github.com/OpenBeta/open-tacos/pull/842

src/db/utils/jobs/migration/ConvertTickDateClimbedFromStringToDate.ts
```diff
@@ -0,0 +1,30 @@
...
+    { dateClimbed: { $exists: true } },
+    {
+      $set: {
+        dateClimbed: {
+          $dateFromString: {
+            dateString: '$dateClimbed',
+            // We want to ascribe an hour of day to the climb, so it shows
+            // up on the correct day when displayed in the user's timezone.
+            // Most climbs are in the US, MT time is a good first approximation.
+            timezone: 'America/Denver'
```
zichongkao: I've not written a Mongo migration before, so please check that this is correct!

vnugent: I think it's ok to assume Mountain time.  This feature will continue to evolve.  Since we have the users MP Id, we can re-import on their behalf, or ask the users to initiate the re-import.

src/db/utils/jobs/migration/__tests__/ConvertTickDateClimbedToDate.ts
```diff
@@ -0,0 +1,58 @@
...
+    await inMemoryDB.clear()
+    ticks = createTickInstance()
+  })
+
+  afterAll(async () => {
+    await server.stop()
+    await inMemoryDB.close()
+  })
+
+  // This will throw an error due to multiple Mongoose connections.
```
zichongkao, line 31: If you know a workaround here, please let me know! I'm not familiar with how these `connectDB` migrations work. Ideally we'd rewired it so we can run the migration functions decoupled from the Mongo `onConnect` event.

src/db/utils/jobs/migration/ConvertTickDateClimbedToDate.ts
```diff
@@ -0,0 +1,35 @@
...
+          }
+        }
+      }
+    }
+  ]))
+
+  await gracefulExit()
+}
+
+// This makes ../__test__/ConvertTickDateClimbedToDate.ts error when importing this file.
```
zichongkao, line 34: If you know a workaround here, please let me know! I'm not familiar with how these connectDB migrations work. Ideally we'd rewired it so we can run the migration functions decoupled from the Mongo onConnect event.

src/utils/inMemoryDB.ts
```diff
@@ -9,18 +9,19 @@ import { logger } from '../logger.js'
  * More portable than requiring user to set up Mongo in a background Docker process.
  * Need a replset to faciliate transactions.
  */
-const mongod = await MongoMemoryReplSet.create({
-  // Stream listener listens on DB denoted by 'MONGO_DBNAME' env var.
-  replSet: { count: 1, storageEngine: 'wiredTiger', dbName: checkVar('MONGO_DBNAME') }
-})
+let mongod: MongoMemoryReplSet
```
zichongkao, line 12: Moving this into `connect` fixes a long-standing bug where jest tests don't exit within a second of finishing due to `openHandles`. When we import these tests, we were creating Mongo connections that were not necessarily closed later on. Now, we only create them on connection, so there is less inadvertent creation.

src/utils/inMemoryDB.ts
```diff
@@ -9,18 +9,19 @@ import { logger } from '../logger.js'
...
  * Connect to the in-memory database.
  */
 const connect = async (): Promise<void> => {
+  mongod = await MongoMemoryReplSet.create({
+    // Stream listener listens on DB denoted by 'MONGO_DBNAME' env var.
+    replSet: { count: 1, storageEngine: 'wiredTiger', dbName: checkVar('MONGO_DBNAME') }
+  })
   const uri = await mongod.getUri(checkVar('MONGO_DBNAME'))
-  logger.info(`Connecting to database ${uri}`)
+  logger.info(`Connecting to in-memory database ${uri}`)
```
zichongkao, line 24: To disambiguate from the other tests which rely on the Docker-initiated, disk-based Mongo instance.

src/db/utils/jobs/migration/ConvertTickDateClimbedToDate.ts
zichongkao: The error is a duplicate Mongoose connection error. Because our test already creates a Mongoose connection to the in-memory Mongo instance. But calling `connectDB` tries to instantiate yet another Mongoose connection.

musoke:
It looks like the date is a required field?

When I surveyed the variety of possible tick sources a few weeks back I found that date is an optional field in some (sendage, possible another).  Imports from biological memory or hand written notes are also likely to be missing dates.

You may have considered this already.

vnugent:
> When I surveyed the variety of possible tick sources a few weeks back I found that date is an optional field in some (sendage, possible another). Imports from biological memory or hand written notes are also likely to be missing dates.

@zichongkao do you want to address this now?  Or let's sit down and discuss an entirely new schema that can support multiple data sources?

zichongkao:
I'm happy for us to sit down and discuss this afresh! I agree the data
structure isn't quite right. Eg. UserID shouldn't be a graphql input since
it can be inferred from middleware, and maybe in general we should be
saving original strings and the inferred version of the strings, eg grades,
climbids, etc.

vnugent:
I ran schema changes, db migration scripts directly with the mongo client.  Can you create a new script here?

https://github.com/OpenBeta/openbeta-graphql/tree/develop/db-migrations

I created utils/jobs/migration for the image migration because the work is a trivial schema change or a db update.  I'm planning to remove `utils/jobs/migration`.

vnugent:
@zichongkao not sure if you saw my comment above.

zichongkao:
Oops yeah I'll get on it soon!

zichongkao:
> @zichongkao not sure if you saw my comment above.

https://github.com/OpenBeta/openbeta-graphql/pull/306 should address it!

PR #302, opened 2023-05-26
Fix up typing in TickDataSource

zichongkao:
Mostly for housekeeping, but sets us up for the migration work we're doing with dateClimbed


src/model/TickDataSource.ts
```diff
@@ -1,47 +1,34 @@
...
   tickModel = getTickModel()
 
   /**
-     * @param tick
-     * takes in a new tick
-     * @returns
-     * returns that new tick
-     */
-  async addTick (tick: TickType): Promise<any> {
-    if (tick === undefined || tick === null) {
```
zichongkao, line 15: These checks are all irrelevant because `tick` is enforced to be non-null by typescript.

PR #308, opened 2023-06-05
DateClimbed is nullable

zichongkao:
We found 20 existing ticks with `dateClimbed` as '0000-00-00'. Since we don't know how to parse these, we are converting them to nulls.

### DIFF
--- a/src/db/TickSchema.ts
+++ b/src/db/TickSchema.ts
@@ -18,7 +18,7 @@ export const TickSchema = new Schema<TickType>({
   userId: { type: Schema.Types.String, required: true, index: true },
   style: { type: Schema.Types.String, required: true, default: '' },
   attemptType: { type: Schema.Types.String, required: true, index: true, default: '' },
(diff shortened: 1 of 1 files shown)


src/db/TickSchema.ts
```diff
@@ -18,7 +18,7 @@ export const TickSchema = new Schema<TickType>({
   userId: { type: Schema.Types.String, required: true, index: true },
   style: { type: Schema.Types.String, required: true, default: '' },
   attemptType: { type: Schema.Types.String, required: true, index: true, default: '' },
-  dateClimbed: { type: Schema.Types.Date, required: true },
+  dateClimbed: { type: Schema.Types.Date },
```
zichongkao, line 21: It might be a problem below that `dateClimbed` is used in the index. Not sure if enabling a nullable field in the index will result in poor performance. `TickSchema.index({ climbId: 1, dateClimbed: 1, style: 1, userId: 1, source: 1 }, { unique: true })`

Issue #309, opened 2023-06-06
Create a new database seeding script that includes user profile data


vnugent:
Currently, we run `yarn seed-usa` to populate our local dev database.   The script only creates climbs and crags.  After #305 we will need the Users collection for development work.  

- [ ] Create a db export script for staging
- [x] Sanitize personal information such as name and email
- [x] Create a new seeding script

PR #311, opened 2023-06-09
chore: new seed-db command

vnugent:
Issue #309 

I haven't figured out an automated way to redact emails so for the time being I manually exported the staging db, redacted emails manually and upload the new db file to google storage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,11 +84,11 @@ query Example1 {
 docker compose up -d
 ```
 
-2. Seed the development database
+2. Seed your local database with data from staging environment
 
 ```bash
 yarn install
-yarn seed-db   # Create test data from staging database
+yarn seed-db   # May take a few minutes to download a large databse file
 ```
 
 3. Start the GraphQL server
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ docker compose up -d
 
 ```bash
 yarn install
-yarn seed-db   # May take a few minutes to download a large databse file
+yarn seed-db   # May take a few minutes to download a large database file
 ```
 
 3. Start the GraphQL server
--- a/README.md
+++ b/README.md
@@ -88,8 +88,7 @@ docker compose up -d
 
 ```bash
 yarn install
-yarn refresh-db   # download USA data files locally and import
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -135,23 +134,6 @@ yarn serve
...
-  # Note: this command will drop and recreate the 'areas' and 'climbs' collection.
-  yarn seed-usa
-
-  # Add all countries (except for USA)
-  yarn add-countries
-
-  # Update area statistics (can be rerun as needed)
-  yarn update-stats
-  ```
-
```
musoke, line 154: Are these being deprecated?

musoke (approved):

README.md
vnugent: I only removed them from the README. I'm voting for deprecating/removing them completely.

seed-db.sh
```diff
@@ -0,0 +1,21 @@
+# This script rebuilds your local database with
+# a copy of OpenBeta staging database.
+#
+#!/bin/bash
+
+FILE_NAME="openbeta-stg-db.tar.gz"
+REMOTE_FILE="https://storage.googleapis.com/openbeta-dev-dbs/$FILE_NAME"
+
+echo "Downloading db file(s)..."
+#wget --content-disposition $REMOTE_FILE
```
siman4457, line 10: @vnugent did you mean to comment this line out? `yarn seed-db` was failing since `openbeta-stg-db.tar.gz` wasn't actually getting downloaded.

vnugent: Ooops, good catch.  Please uncomment it.  Better yet, if you can improve the script so that we can either: download and rebuild db, or rebuild only (skip the download because it's a large file)

siman4457: [Uncommented](https://github.com/OpenBeta/openbeta-graphql/commit/dab9388193d4733e577c454a31354245253fdced) for now.

For improving the script:
Should the decision to download or skip download be from user input? 

We can automatically start a download if we detect the tar file is outdated. Can we perform any versioning or date comparison between what's on Google storage and what we have locally?

Issue #281, opened 2023-05-03
Migrate Tag CRUD to new data model


vnugent:
Media & Tags models have changed significantly in PR #273 breaking all Tag operations

PR #312, opened 2023-06-10
Migrate tag api to new media model

vnugent:
Issue #281 
- [x] refactor: migrate tag mutations to work with new media model
- [x] feat: introduce permission checks to tag mutations

Clean up
- [x] refactor: add singleton factory function to data source classes
- [x] chore: remove unused code (I was going to do this in separate PR but the tag api changes touch a lot of other code)


src/model/MutableClimbDataSource.ts
```diff
@@ -275,8 +275,15 @@ export default class MutableClimbDataSource extends ClimbDataSource {
...
     return ret
   }
-}
 
-// Why suppress TS error? See: https://github.com/GraphQLGuide/apollo-datasource-mongodb/issues/88
-// @ts-expect-error
-export const createInstance = (): MutableClimbDataSource => new MutableClimbDataSource(getClimbModel())
+  static instance: MutableClimbDataSource
+
+  static getInstance (): MutableClimbDataSource {
```
vnugent, line 281: a singleton factory

src/auth/permissions.ts
```diff
@@ -13,7 +13,9 @@ const permissions = shield({
     updateArea: isEditor,
     updateClimbs: isEditor,
     deleteClimbs: isEditor,
-    updateUserProfile: and(isOwner, isValidEmail)
+    updateUserProfile: and(isOwner, isValidEmail),
+    addEntityTag: or(isMediaOwner, isUserAdmin),
+    removeEntityTag: or(isMediaOwner, isUserAdmin)
```
vnugent, line 18: tag mutations now require permission checks

src/db/MediaObjectTypes.ts
```diff
@@ -1,11 +1,11 @@
-import { ObjectId } from 'mongoose'
+import mongoose from 'mongoose'
 import { MUUID } from 'uuid-mongodb'
 import { Point } from '@turf/helpers'
 
 export type ImageFormatType = 'jpeg' | 'png' | 'webp' | 'avif'
 
 export interface MediaObject {
-  _id: ObjectId
+  _id: mongoose.Types.ObjectId
```
vnugent, line 8: we've been using ObjectId type incorrectly.  

- `mongoose.Types.ObjectId` is the typescript type and also the class to generate new mongo object id
- `mongoose.ObjectId` should only be used for describing the schema

vnugent (commented):
Reviving media tag API

.vscode/launch.json
```diff
@@ -68,50 +55,10 @@
                 "--inspect-brk",
                 "${workspaceRoot}/node_modules/.bin/jest",
                 "--runInBand",
-                "UserData"
+                "MediaData"
             ],
             "console": "integratedTerminal",
             "internalConsoleOptions": "neverOpen",
-        },
```
vnugent, line 75: removing Visual Studio code debugging configs related to media migration

src/model/MutableMediaDataSource.ts
```diff
@@ -1,119 +1,143 @@
...
+          _id: new mongoose.Types.ObjectId(),
+          targetId: entityUuid,
+          type: entityType,
+          ancestors: area.ancestors,
+          areaName: area.area_name,
+          lnglat: area.metadata.lnglat
+        }
 
-        if (await media.exists(filter) != null) throw new UserInputError('Duplicate mediaUuid and destinationId now allowed')
+        // We treat 'entityTags' like a Set - can't tag the same climb/area id twice.
```
zichongkao: Small thing but this part onward is duplicated in case 0 above. Might be worth moving it out of the switch statement to deduplicate.

src/model/__tests__/MediaDataSource.ts
```diff
@@ -1,143 +1,166 @@
...
+import muuid, { MUUID } from 'uuid-mongodb'
 import MutableMediaDataSource from '../MutableMediaDataSource'
 import AreaDataSource from '../MutableAreaDataSource'
+import ClimbDataSource from '../MutableClimbDataSource'
 
-import { connectDB, createIndexes, getAreaModel, getMediaModel } from '../../db/index.js'
-import { MediaInputType, TagEntryResultType } from '../../db/MediaTypes.js'
+import { connectDB, createIndexes } from '../../db/index.js'
 import { AreaType } from '../../db/AreaTypes.js'
+import { EntityTag, MediaObject, MediaObjectGQLInput, AddTagEntityInput } from '../../db/MediaObjectTypes.js'
```
zichongkao, line 9: I've been favoring writing integration tests so that we can validate and document the graphQL layer as well rather than just data sources down. They are in `src/__tests__/`. We could add tests at both layers too, but I'd suggest writing the integration tests if we're only going to write one.

zichongkao (approved):
Nice! I like the use of the singleton pattern. Should reduce memory usage!

src/model/__tests__/MediaDataSource.ts
vnugent: Agreed. We can also test the permission layer that way.  I've been writing data source tests out of habit.  I'll fix it in a future PR.

src/model/MutableMediaDataSource.ts
vnugent: Good point.  I'll submit a change.

musoke (approved):

Issue #310, opened 2023-06-08
Get user ticks API's input should be username instead of user uuid


vnugent:
Required by FE  https://github.com/OpenBeta/open-tacos/issues/864

https://github.com/OpenBeta/openbeta-graphql/blob/485120b7c7b55c1e9747c4bc52ef9edb7f13ba4e/src/graphql/schema/Tick.gql#L5

PR #313, opened 2023-06-11
Query ticks by username

zichongkao:
In response to https://github.com/OpenBeta/openbeta-graphql/issues/310.

Rather than switch out the API endpoint as the task asks, I think it might still be useful to be able to query ticks by userID, so I chose to preserve the functionality.

To do this, I could create a brand new endpoint to query by username. But I figure the userTick endpoint can be simply generalized so that a user's ticks can be retrieved by any number of the user's identifiers. This means that we need one more Mongo query to search for the user object. However, this allows us to validate that the userId supplied does exist. Otherwise if the endpoint returns an empty array, the frontend cannot differentiate if there is no such userId or if the user has no ticks.

### DIFF
--- a/src/model/TickDataSource.ts
+++ b/src/model/TickDataSource.ts
@@ -2,11 +2,14 @@ import { MongoDataSource } from 'apollo-datasource-mongodb'
 import type { DeleteResult } from 'mongodb'
 import mongoose from 'mongoose'
 
-import { TickEditFilterType, TickInput, TickType } from '../db/TickTypes'
-import { getTickModel } from '../db/index.js'
+import { TickEditFilterType, TickInput, TickType, TickUserSelectors } from '../db/TickTypes'
+import { getTickModel, getUserModel } from '../db/index.js'
+import { canonicalizeUsername } from '../utils/helpers'
+import type { User } from '../db/UserTypes'
 
 export default class TickDataSource extends MongoDataSource<TickType> {
   tickModel = getTickModel()
+  userModel = getUserModel()
 
   /**
    * @param tick takes in a new tick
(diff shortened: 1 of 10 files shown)


src/model/TickDataSource.ts
```diff
@@ -71,11 +74,39 @@ export default class TickDataSource extends MongoDataSource<TickType> {
...
+    const { userId: requestedUserId, username } = userSelectors
+    if (requestedUserId == null && username == null) {
+      throw new Error('Username or userId must be supplied')
+    }
+    const filters: any[] = []
+    if (requestedUserId != null) {
+      filters.push({ _id: requestedUserId })
+    }
+    if (username != null) {
+      filters.push({ // Fuzzy match based on canonicalName since those are unique to a user.
```
zichongkao: The other reason I thought we should filter using `canonicalName` is because we have an index on it but not on `username`.

vnugent: Not having an index on username was an oversight since it's an important query for the profile page. 
The 'fuzzy' match was only intended for checking username uniqueness when updating/creating username.

To simplify things I don't think we should support fetching ticks by canonical name.  What do you think?
We can look into supporting a true typo-tolerant username matching in the future if there's a use case for it.

vnugent (approved):

vnugent:
Is it possible to merge https://github.com/OpenBeta/openbeta-graphql/pull/312 first and then rebase yours?

I've deleted a lot of unused migration code and added a factory function to all data source classes.

src/model/TickDataSource.ts
zichongkao: Sounds good. I can fix it so that it searches strictly by username when I rebase on top of #312 !

vnugent:
deployed to stg and prod.

Issue #213, opened 2023-01-01
Breadcrumbs not updated when users updating area name


vnugent:
When an area is created, we also generate its bread crumbs data in `pathTokens`  field.  However, when we update an area's name, we don't update `pathTokens`.

![openbeta-breadcrumbs-bug](https://user-images.githubusercontent.com/3805254/210170501-9335d965-5039-447a-9a8a-72d4452695c7.jpeg)

### To do:
- Case 1:  The area has no child areas.  Simply update `pathTokens` when we update area's name.
- Case2: The area has child areas.  We need to update all of its children's pathTokens

CocoisBuggy:
I can fix this but I just want to make absolutely sure we don't want to resolve this at query time rather than at mod time?

We chatted about reads being far more frequent than writes / modifications but I just want to make sure that the pathTokens system is still the way we want to go?

vnugent:
> I can fix this but I just want to make absolutely sure we don't want to resolve this at query time rather than at mod time?
> 
> We chatted about reads being far more frequent than writes / modifications but I just want to make sure that the pathTokens system is still the way we want to go?

Correct.  We need to update `pathToken`, either iteratively or recursively, when doing `updateArea()`,

saferthanhouses:
@vnugent Happy to look into this if it's available.

zichongkao:
@saferthanhouses @CocoisBuggy I know it's been a while since we've updated this issue. Are either of you still open to taking it on?

andrew-jp:
I'm new here/have no idea what I'm doing, but I've been working on this all day. What do you guys think of the general idea? 

``` javascript
  updateArea()...
  ...
      const area = await this.areaModel.findOne(filter).session(session)
      if (area == null) {
        throw new Error('Area update error.  Reason: Area not found.')
      }

      const { areaName, description, shortCode, isDestination, isLeaf, isBoulder, lat, lng, experimentalAuthor } = document

  ...
      if (areaName != null) {
        const sanitizedName = sanitizeStrict(areaName)
        area.set({ area_name: sanitizedName })
        area.set({ pathTokens: area.pathTokens.slice(0, -1).push(sanitizedName) })

        if (area.children.length > 0) {
          await this.updatePathTokens(area.children, sanitizedName)
        }
      }

    ...
  async updatePathTokens (areaChildren: Types.ObjectId[], areaName: string, index: number = 2): Promise<void> {
    const session = await this.areaModel.startSession()

[...]

PR #314, opened 2023-06-12
Breadcrumbs issue #213 fix

andrew-jp:
Issue #213 

This is a working fix for the breadcrumbs issue. It copies/updates/sets the updated areas tokens, and then calls a recursive function on each of the areas children (in this case just Mongoose object id's). The recursive function looks up the object id (queries the db), and if it has path tokens it updates the index (a depth counter is passed into each recursive call to track this index) where the name change occurred. It then sets the pathTokens in the object and saves. It will throw if the save fails or if the query for the child object returns null.

Performance varied a little in my testing. To update a large area (Central Wasatch in Utah for example) it needs to update a large tree structure, with all the associated query and save ops. This initially took about 7.5s, but in my recent testing only about 5s. I don't how much this is machine dependent, but I imagine this solution isn't performant regardless. I'll try to improve the performance where I can.

Any input on error handling and performance is welcome.

### DIFF
--- a/src/model/MutableAreaDataSource.ts
+++ b/src/model/MutableAreaDataSource.ts
@@ -1,7 +1,7 @@
 import { geometry, Point } from '@turf/helpers'
 import muuid, { MUUID } from 'uuid-mongodb'
 import { v5 as uuidv5, NIL } from 'uuid'
-import mongoose, { ClientSession } from 'mongoose'
+import mongoose, { ClientSession, Types } from 'mongoose'
 import { produce } from 'immer'
 import { UserInputError } from 'apollo-server'
 import isoCountries from 'i18n-iso-countries'
@@ -301,7 +301,20 @@ export default class MutableAreaDataSource extends AreaDataSource {
         throw new Error('Area update error.  Reason: Updating leaf or boulder status of an area with subareas is not allowed.')
       }
 
-      if (areaName != null) area.set({ area_name: sanitizeStrict(areaName) })
+      if (areaName != null) {
+        const sanitizedName = sanitizeStrict(areaName)
+        area.set({ area_name: sanitizedName })
+
+        const newPath = [...area.pathTokens]
+        newPath.pop()
+        newPath.push(sanitizedName)
+        area.set({ pathTokens: newPath })
+
+        for (const childId of area.children) {
+          await this.updatePathTokens(childId, sanitizedName)
+        }
+      }
+      
       if (shortCode != null) area.set({ shortCode: shortCode.toUpperCase() })
       if (isDestination != null) area.set({ 'metadata.isDestination': isDestination })
       if (isLeaf != null) area.set({ 'metadata.leaf': isLeaf })
@@ -356,6 +369,34 @@ export default class MutableAreaDataSource extends AreaDataSource {
         return ret
       })
     return ret
+  }
+
+  async updatePathTokens (childId: Types.ObjectId, newAreaName: string, index: number = 2): Promise<void> {
+    // const session = await this.areaModel.startSession()
+    const filter = { _id: childId }
+    // const area = await this.areaModel.findOne(filter).session(session)
+    const area = await this.areaModel.findOne(filter)
+
+    if (area == null) {
+      throw new Error('pathTokens update error.  Reason: child area not found.')
+    }
+
+    if (area.pathTokens.length > 0) {
+      const newPath = [...area.pathTokens]
+      newPath[newPath.length - index] = newAreaName
+      area.set({ pathTokens: newPath })
+      area.save(function (err, result) {
+        if (err != null) {
+          throw new Error('pathTokens update error.  Reason: save operation failed.')
+        } else {
+          console.log(result)
+        }
+      })
+
+      for (const childId of area.children) {
+        await this.updatePathTokens(childId, newAreaName, index + 1)
+      }
+    }
   }
 
   /**
--- a/src/model/MutableAreaDataSource.ts
+++ b/src/model/MutableAreaDataSource.ts
@@ -1,7 +1,7 @@
 import { geometry, Point } from '@turf/helpers'
 import muuid, { MUUID } from 'uuid-mongodb'
 import { v5 as uuidv5, NIL } from 'uuid'
-import mongoose, { ClientSession, Types } from 'mongoose'
+import mongoose, { ClientSession } from 'mongoose'
 import { produce } from 'immer'
 import { UserInputError } from 'apollo-server'
 import isoCountries from 'i18n-iso-countries'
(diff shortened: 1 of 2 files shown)


src/model/MutableAreaDataSource.ts
```diff
@@ -301,7 +301,22 @@ export default class MutableAreaDataSource extends AreaDataSource {
...
+
+        // change our pathTokens
+        const newPath = [...area.pathTokens]
+        newPath.pop()
+        newPath.push(sanitizedName)
+        area.set({ pathTokens: newPath })
+
+        // iterate over the DB id's in the child list and update pathTokens
+        for (const childId of area.children) {
+          await this.updatePathTokens(childId, sanitizedName)
```
vnugent: You should pass the `session` variable down to `updatePathTokens()` so that an exception will rollback everything.

vnugent (changes requested):

src/model/MutableAreaDataSource.ts
andrew-jp: I believe I fixed this - I'm not sure how to get 'save' to throw - but I threw a generic error in the same scope a few levels into the recursion and it canceled the save operation on the site, and passed the error to the on screen error notification.

vnugent:
@andrew-jp I checked out the PR locally and did a quick smoke test.   It updated all nested child areas as expected and took less than 3s on my old Macbook.  Nice work!

I notice the change history code generates a lot of warnings, "History Id not found.  Ignore change."  I think we will also need to pass down the history Id so that all recursive updates are also recorded in the same changeset.  I'll dig a bit deeper and let you know. 

Meanwhile can you look into adding a few unit tests? https://github.com/OpenBeta/openbeta-graphql/blob/develop/src/model/__tests__/updateAreas.ts

src/model/MutableAreaDataSource.ts
```diff
@@ -301,7 +301,21 @@ export default class MutableAreaDataSource extends AreaDataSource {
...
+        const sanitizedName = sanitizeStrict(areaName)
+        area.set({ area_name: sanitizedName })
+
+        // change our pathTokens
+        const newPath = [...area.pathTokens]
+        newPath[newPath.length - 1] = sanitizedName
+        area.set({ pathTokens: newPath })
+
+        // iterate over the DB id's in the child list and update pathTokens
+        for (const childId of area.children) {
```
vnugent: You may be able to eliminate the for loop and multiple find() calls by either:

- Calling mongoose populate() function to turn the array of area IDs into array of area objects.
- Using the $in operator to get all child area objects in 1 call

Similar code
https://github.com/OpenBeta/openbeta-graphql/blob/develop/src/db/utils/jobs/TreeUpdater.ts#L47

This is just a comment, not a request to change.  If you want, you can do performance optimization in a follow up PR.

vnugent: I've pushed a small modification to `updatePathTokens()`, calling `populate()` to hydrate area.children[] to avoid calling find() in a for loop. I renamed Oregon state and saw an improvement from 5s down to 2s.

src/model/MutableAreaDataSource.ts
```diff
@@ -293,6 +295,25 @@ export default class MutableAreaDataSource extends AreaDataSource {
...
+      // See https://github.com/OpenBeta/openbeta-graphql/issues/244
+      let experimentaAuthorId: MUUID | null = null
+      if (experimentalAuthor != null) {
+        experimentaAuthorId = await this.experimentalUserDataSource.updateUser(session, experimentalAuthor.displayName, experimentalAuthor.url)
+      }
+
+      const opType = OperationType.updateArea
+      const change = await changelogDataSource.create(session, user, opType)
+
+      const _change: ChangeRecordMetadataType = {
```
vnugent, line 307: move _change object up so that we can pass it down to updatePathTokens()

andrew-jp: Glad it worked for you as well, I'll see what I can do about adding some unit tests.

andrew-jp:
I added a basic test that checks a small tree's pathTokens before/after the update. It matches them exactly, ie: ['Japan', 'Test Change', 'B1', 'C1']. I'd like to have a test that verifies a total rollback if we get an error in the recursion, but I don't know how to go about creating that error. I'll consider more cases tomorrow.

vnugent:
Not 100% sure if it's going to work.  You can try mocking the implementation of `updatePathTokens()`

On the first invocation return the default implementation.  On the 2nd (or 3rd) mock the implementation with a function that simply throws an exception.

https://stackoverflow.com/questions/61450440/jest-spyon-to-mock-implementation-only-on-second-call-and-the-third-call
https://stackoverflow.com/questions/62225496/modify-mockimplementation-results-in-spyon-of-jest

src/model/MutableAreaDataSource.ts
```diff
@@ -376,7 +376,7 @@ export default class MutableAreaDataSource extends AreaDataSource {
    * @param newAreaName new area name
    * @param depth tree depth
    */
-  async updatePathTokens (session: ClientSession, changeRecord: ChangeRecordMetadataType, area: AreaDocumnent, newAreaName: string, depth: number = 2): Promise<void> {
+  async updatePathTokens (session: ClientSession, changeRecord: ChangeRecordMetadataType, area: AreaDocumnent, newAreaName: string, depth: number = 1): Promise<void> {
     if (area.pathTokens.length > 1) {
```
vnugent, line 380: I changed starting depth to 2 because we don't allow updating country name which has pathTokens.length === 1.  Is it causing an issue?

https://github.com/OpenBeta/openbeta-graphql/blob/2d79c70100a42a37c083ed575fd7d0d0dddbeb3d/src/model/MutableAreaDataSource.ts#L297

andrew-jp: The last checks showed some weird test results where it looked like we were using the wrong index. Since the first iteration updates the last item in the areas pathTokens, it should be [ pathTokens.length - 1 ] right? As it was, it was starting at [ length - 2 ]. I tested it locally before and after and was having issues with the 2.

ex: update Wasatch 
pathTokens = [USA, Utah, Wasatch]
newPath = [USA, Utah, Wasatch]
newPath[ (3 - 2) ] = New Name
newPath = [USA, New Name, Wasatch]

andrew-jp: I just realized instead of passing depth down through the calls, we just have to know the index of the initial update and pass that index down to the children - in the case above, the index of Wasatch would be 2 in all the childrens pathTokens.

andrew-jp: I just realized that I don't need to index from the back because the result of (length - depth) is constant. I just overthought this initially.

ex: Index of Wasatch is 2 in self and children
[USA, Utah, Wasatch]
[USA, Utah, Wasatch, Little Cottonwood]
[USA, Utah, Wasatch, Little Cottonwood, The Thumb]

vnugent:
We can add more unit tests in a follow up PR

musoke:
I didn't notice this got merged until I saw it in action on openbeta.io - nicely done @andrew-jp!

> We can add more unit tests in a follow up PR
Do we need an issue this or are they low priority?

Issue #320, opened 2023-06-19
User media & tags query need pagination support


vnugent:
Needed by Frontend https://github.com/OpenBeta/open-tacos/pull/869 

### Background (TBD)

PR #322, opened 2023-06-20
feat: add pagination support to user media query

vnugent:
Issue #320 
Needed by frontend PR: https://github.com/OpenBeta/open-tacos/pull/869

I followed [GQL recommended approach](https://graphql.org/learn/pagination/) for implementing a cursor-based pagination.

```gql
# Sample query
query GetNextPage {
  getUserMediaPagination(
    input: {userUuid: "b9f8ab3b-e6e5-4467-9adb-65d91c7ebe7c", first: 3, after: "1675245914389_645aa64261c73112fc19b505"}
  ) {
    userUuid
    mediaConnection {
      edges {
        cursor
        node {
          username
          mediaUrl
          entityTags {
            climbName
            id
          }
          id
        }
      }
    }
  }
}
```

### DIFF
--- a/src/model/MediaDataSource.ts
+++ b/src/model/MediaDataSource.ts
@@ -1,9 +1,9 @@
 import { MongoDataSource } from 'apollo-datasource-mongodb'
 import muid, { MUUID } from 'uuid-mongodb'
-import mongoose from 'mongoose'
+import mongoose, { ObjectId } from 'mongoose'
 import { logger } from '../logger.js'
 import { getMediaObjectModel } from '../db/index.js'
-import { TagsLeaderboardType, AllTimeTagStats, MediaByUsers, MediaForFeedInput, MediaObject } from '../db/MediaObjectTypes.js'
+import { TagsLeaderboardType, UserMediaQueryInput, AllTimeTagStats, MediaByUsers, MediaForFeedInput, MediaObject } from '../db/MediaObjectTypes.js'
 
 const HARD_MAX_FILES = 1000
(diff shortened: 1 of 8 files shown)


src/model/MediaDataSource.ts
```diff
@@ -20,6 +20,19 @@ export default class MediaDataSource extends MongoDataSource<MediaObject> {
...
+   * Find one media object by id.  Throw an exception if not found.
+   * @param _id
+   */
+  async getOneMediaObjectById (_id: mongoose.Types.ObjectId): Promise<MediaObject> {
+    const rs = await this.mediaObjectModel.find({ _id }).orFail(new Error('Media not found')).lean()
+    if (rs != null && rs.length === 1) {
+      return rs[0]
+    }
+    logger.error(`This shouldn't happend. Found multiple media objects for id: ${_id.toString()}`)
+    throw new Error('Media not found for id')
```
musoke, line 33: Is this error message sufficiently illuminating?  Makes it sound like 0 foumd, not >1

musoke (approved):
I left one comment on something I may have misunderstood - other than that it looks good to me as a non-expert.

src/model/MediaDataSource.ts
vnugent: I intentionally make the message slightly different than the case where there is 0 so we can debug the issue. And at the time, make it a bit vague in order to not scare users that we have corrupted data.  Since media._id is unique theoretically we should never get back more than 1 document.

musoke:
Ok, sounds good!

PR #340, opened 2023-07-13
fix: Allow identical ticks on one day

### LINKED ISSUE #631: MP import feature only shows 3
## Expected Behavior
MP user: https://www.mountainproject.com/user/200492669/andre-chiquito
OB profile: https://openbeta.io/u/all_terrain_human

The user has over 400 ticks but it seems only 3 are imported/shown

cc @Downster

musoke:
The schema enforces a unique index on {climbId, dateClimbed, style, userID, source}.  This is meant to prevent duplicating ticks imported from mountain project.  However, it also prevents logging such ticks when they are legitimate.

The import procedure also deletes imported ticks before reimporting, so this was overly cautious.

Change the schema so it does not enforce uniqueness in this way. Instead, add non-unique indexes on { userId } and { userId, climbId }.

Add a db migration script.

Remove climb uniqueness test.

Fixes https://github.com/OpenBeta/open-tacos/issues/631

### DIFF
--- a/db-migrations/0004-unique-user-climb-date-style.js
+++ b/db-migrations/0004-unique-user-climb-date-style.js
@@ -1,0 +1,11 @@
+/**
+ * https://github.com/OpenBeta/open-tacos/issues/631
+ **/
+
+rs1 = db.ticks.createIndex({ userId: -1 })
+rs2 = db.ticks.createIndex({ userId: -1, climbId: -1 })
+rs3 = db.ticks.dropIndex({ climbId: 1, dateClimbed: 1, style: 1, userId: 1, source: 1 })
+
+printjson(rs1)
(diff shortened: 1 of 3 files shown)


musoke:
I tried a couple known-problematic logbooks and each worked as expected.

vnugent (approved):

vnugent:
Deployed to staging and production

musoke:
Did you run the migration?  I haven't written one before so if you haven't done it yet, double check the script before you use it!

vnugent:
> Did you run the migration? I haven't written one before so if you haven't done it yet, double check the script before you use it!

Done.  Can you try importing again?

musoke:
It works!

Issue #266, opened 2023-04-19
 multipitch routes 


musoke:
Is there a plan for how multi-pitch routes will be handled?  I don't see much documentation of this.

#252 has some discussion.  The suggestion there is that multipitch routes should get an array of bolt counts, one for each pitch.

Presumably the same would work for grades.  Sandbag grades are ordered, so the overall route grade can be calculated as the maximum over the pitches.  There will be more subtlety when different pitches have different disciplines (e.g. free and ice on one climb).

What other fields need generalizing?

@bradleyDean asks about variations in #252. Simplest suggestion is to list them as separate routes.

l4u532:
As in https://github.com/OpenBeta/openbeta-graphql/pull/252 I suggest

- adding new field "bolts" with String array that contains no of bolts per pitch
- changing "grade" to String array to include difficulty per pitch
- (and route variants are represented as individual routes, this preserves the option to tick routes)

This should be a user-friendly solution to represent multipitches.

musoke:
> * (and route variants are represented as individual routes, this preserves the option to tick routes)

This is certainly the simplest and most obvious solution.  It might get messy in practice though.  What if I climb the bottom half of one route and link it into the top half of another?  Then I would need to add another route, even though all its pitches already exist elsewhere in the database.

We could instead allow for ticks which mix and match pitches from different routes.  Then variations can be listed without duplicating pitches and the user can still tick what they actually climbed.

l4u532:
I guess in that case one multi-pitch route could have several sub-routes, with individual difficulty, bolts, description. Sounds neat. Yet, I have to admit, viewed from my central European climbing perspective, ticking mixed pitches would be a niche feature (the vast majority of climbs are bolted here).

If you don't disagree with points (1) and (2), I'd try implementing them, as I/we need bolt count for a sensible representation in Central Europe and (3) should not be affected by a String array implementation as proposed (1)+(2).

musoke:
Sounds good!

To be clear: when I said "mix" I meant combining pitches from different routes, not mixed protection.  I hadn't thought about that.

vnugent:
Why would you use string instead of a valid grade that passes `openbeta/sandbag` validation?

> changing "grade" to String array to include difficulty per pitch

l4u532:
Maybe a misunderstanding (I am aware of `Climb.grades`). Anyway, I have been thinking about this and have come to a different conclusion altogether - I am trying to recap the problem and options:

[...]

PR #350, opened 2023-08-05
feat: add multipitch support

l4u532:
# Problem

(See https://github.com/OpenBeta/openbeta-graphql/issues/266)

For a given multi-pitch climb, for each pitch, we want to represent multiple boltsCount and grades, e. g. grade = ["6a", "6c", "5c", "6c"], and boltsCount = [4, 6, 3, 3], and possible other data (like description).

# Solution

Introduce type `Pitch` (with fields `parentId`, `type`, `number`, `grade`, `length`, `boltsCount`, `description`) as a child to `climb`, so that `climb` gets an optional Array `Climb.pitches` containing each pitch; this has the most flexibility and allows for detailed per-pitch data, plus it wouldn't break any existing implementation

## Example JSON Doc Schema
```
type Climb {
  id: ID!
  uuid: ID!
  name: String!
  ...
  pitches: [Pitch]!  # new pitches array of type Pitch
}

```

The Type `Pitch` would contain:

```
type Pitch {
  uuid: ID!
  parentId: ID!
  number: Int!
  grades: GradeType!
  type: ClimbType!
  length: Int!
  boltsCount: Int
  description: String
}

```


l4u532:
@musoke Would you kindly have a look? Thank you =)

vnugent:
@l4u532 I'll have a look at this issue by this weekend

src/db/ClimbSchema.ts
```diff
@@ -59,6 +59,28 @@ const MetadataSchema = new Schema<IClimbMetadata>({
...
+  _id: {
+    type: 'object',
+    value: { type: 'Buffer' },
+    default: () => muuid.v4()
+  },
+  uuid: {
+    type: 'string',
+    default: function () { return this._id.toString() }
+  },
+  parentId: { type: String, required: true },
```
musoke: Is this the id of the route that the pitch is part of?  Do we allow a pitch to be part of multiple routes?

src/db/ClimbSchema.ts
```diff
@@ -59,6 +59,28 @@ const MetadataSchema = new Schema<IClimbMetadata>({
...
+  number: { type: Number, required: true },
```
musoke: What is number?  Is this to track the order of the pitches?

musoke: Looks like it from the rest of the changes.

src/model/__tests__/MutableClimbDataSource.ts
```diff
@@ -109,6 +109,36 @@ describe('Climb CRUD', () => {
     grade: 'WI8+'
   }
 
+  // Define a sport climb with two individual pitches
+  const newClimbWithPitches: ClimbChangeInputType = {
+    name: 'Short Multi-Pitch',
+    disciplines: {
+      sport: true
+    },
+    grade: '7', // max grade of its child pitches
```
musoke, line 118: Does the grade for the whole route need to be stated explicitly?  Should sandbag be able to convert a list of grades of pitches to an overall grade?  (Something like `reduce_multipitch([6+, 7] = 7`)

This could wait until this initial version of multipitch routes is complete.

musoke (commented):
I had to reread #266 to remind myself of all the nuances discussed previously.  This looks good, although I haven't run it myself. 

We discussed allowing a pitch to be part of multiple routes, to account for variations, etc.  Does this PR allow for that to be added in future? The `number` and `parent` fields assume only one parent route.

src/model/__tests__/MutableClimbDataSource.ts
l4u532: `grades` is not a mandatory field. And absolutely, I had such a reduce-logic in mind, too. However, I wanted to keep v1 as simple as possible.

https://github.com/OpenBeta/openbeta-graphql/blob/3ecbd178ce1e35b90a1110e4ad3f8de926315520/src/graphql/schema/Climb.gql#L26

src/db/ClimbSchema.ts
l4u532: `number` is to track the sequence of pitches. It is also only required field besides `id` and `parentId`. The reasoning: the correct sequence of information is essential for multi-pitches. A user might only add the pitches' `grades`; but it is prime that she does so in correct order, else the value of information is almost zero.

Since this seems to be non-intuitive, should I rename this field for more clarity? E. g. `pitchNumber`? (Suggestions welcome)

src/db/ClimbSchema.ts
l4u532: This implementation allows for *ticking* variations (since each pitch gets a UUID), but not for having multiple parent `climb`s for one `pitch`. I think it is preferable to leave it out because it would reduce clarity a lot in return for what I believe is an unfrequent requirement. Let me explain:

The implementation would require a complimentary field to track the pitches' sequence by `pitch.id`, e. g. `Climb.pitchSequence`: [$id1, $id2, $id3, $id4, ...]. This would either lie outside the `Pitch` object, or make the `Pitch` object nested (e. g. `Pitch.pitchSequence`and `Pitch.data.x`.

Granted, it would indeed be useful for climbing areas with limited space or for mixed climbing (trad/sport, mixed (ice+drytooling)). But I feel like duplicating a pitch here and then seems more feasible. This is based on my ballpark estimate that variations only occur for 5-10% of multi-pitch routes.

Eager to hear your thoughts on this :-)

l4u532:
> I had to reread #266 to remind myself of all the nuances discussed previously. This looks good, although I haven't run it myself.
> 
> We discussed allowing a pitch to be part of multiple routes, to account for variations, etc. Does this PR allow for that to be added in future? The `number` and `parent` fields assume only one parent route.

See https://github.com/OpenBeta/openbeta-graphql/pull/350#discussion_r1309774943 for my reply.

src/model/__tests__/MutableClimbDataSource.ts
musoke: Great!

src/db/ClimbSchema.ts
```diff
@@ -59,6 +59,28 @@ const MetadataSchema = new Schema<IClimbMetadata>({
   }
 }, { _id: false })
 
+const PitchSchema = new mongoose.Schema({
+  _id: {
+    type: 'object',
+    value: { type: 'Buffer' },
+    default: () => muuid.v4()
+  },
+  uuid: {
```
vnugent: is `uuid` field redundant since we can always convert _id to UUID format?

src/db/ClimbSchema.ts
vnugent: We should use something else besides `number` to avoid conflict with `number` in Typescript. `pitchNumber` sounds good.  At first I thought pitch number isn't needed since we can get them from the array index.  But in some case it may not be easy to get the index.  Glad you thought of it.

src/db/ClimbSchema.ts
```diff
@@ -59,6 +59,28 @@ const MetadataSchema = new Schema<IClimbMetadata>({
...
+    type: 'string',
+    default: function () { return this._id.toString() }
+  },
+  parentId: { type: String, required: true },
+  number: { type: Number, required: true },
+  grades: { type: mongoose.Schema.Types.Mixed },
```
vnugent: Is there a good reason not to use the `GradeTypeSchema`?

src/db/ClimbSchema.ts
```diff
@@ -59,6 +59,28 @@ const MetadataSchema = new Schema<IClimbMetadata>({
...
+  type: { type: mongoose.Schema.Types.Mixed },
+  length: { type: Number },
+  boltsCount: { type: Number },
+  description: { type: String }
+}, {
+  _id: true,
```
vnugent: we should remove **`_id: true`** since we will be providing our own UUID.  `_id: true` means Mongo will auto-generate _id for us in ObjectID format (not UUID).

src/db/ClimbSchema.ts
vnugent: Since `parentID` references Climb._id, the type should be `Schema.Types.Mixed`.  You can see an example in Area.climbs[] Schema.  https://github.com/OpenBeta/openbeta-graphql/blob/4ce0c36a4941e1654c24fd80aac901555489a2cf/src/db/AreaSchema.ts#L94

vnugent (changes requested):
@l4u532 Thank you very much for this PR, and I apologize for my delayed response.  I left my feedback inline.

src/db/ClimbSchema.ts
l4u532: @vnugent I have incorporated all changes and would be happy to get your re-review, especially on lines 77-81, since I am not sure if I have interpreted your suggestions correctly: https://github.com/OpenBeta/openbeta-graphql/pull/350/files#diff-9b4028942ce1d106899ff51138cc6ee7cc75ea36110257394c51424643261967R77-R81

src/db/ClimbTypes.ts
```diff
@@ -27,21 +27,35 @@ export type ClimbGQLQueryType = ClimbType & {
...
 export type ClimbType = IClimbProps & {
+  pitches?: IPitch[]
   metadata: IClimbMetadata
   content?: IClimbContent
 }
 
+/* Models a single pitch of a multi-pitch route */
+export interface IPitch {
+  _id: MUUID
+  parentId: string
```
vnugent: Can we change `IPitch.parentId` to MUUID?  Since it references `Climb._id` which has MUUID type.

src/graphql/resolvers.ts
```diff
@@ -153,6 +153,18 @@ const resolvers = {
 
     boltsCount: (node: ClimbGQLQueryType) => node.boltsCount ?? -1,
 
+    pitches: (node: ClimbGQLQueryType) => node.pitches != null
+      ? node.pitches.map(pitch => {
+        const { parentId, ...otherPitchProps } = pitch
+        return {
+          id: pitch._id?.toUUID().toString(),
+          uuid: pitch._id?.toUUID().toString(),
```
vnugent: what do you think if we just expose `id` and remove `uuid`? (since they both serve the same purpose)

src/model/MutableClimbDataSource.ts
```diff
@@ -125,6 +125,23 @@ export default class MutableClimbDataSource extends ClimbDataSource {
...
+      const newPitchesWithIDs = pitches != null
+        ? pitches.map((pitch): IPitch => {
+          if (pitch.pitchNumber === undefined) {
+            throw new UserInputError('Each pitch in a multi-pitch climb must have a pitchNumber representing its sequence in the climb. Please ensure that every pitch is numbered.')
+          }
+
+          return {
+            ...pitch,
+            _id: muid.from(pitch.id ?? muid.v4()), // generate MUUID if not present
+            parentId: muid.from(pitch.parentId ?? newClimbIds[i]).toString(),
```
vnugent: Per my feedback above about changing IPitch.parentId to MUUID, we can just assign `newClimbIds[i]` to parentId (not doing toString())

vnugent (changes requested):
Thanks for the updating the code.  I think we're very close now.  I left some comments inline.

src/graphql/resolvers.ts
l4u532: Sure. I was copying it from https://github.com/l4u532/openbeta-graphql/blob/4fd8509f79fd59c04cc39430a2c01ad2d0f8dfe2/src/graphql/resolvers.ts#L134C1, I assumed it might have a specific motivation to expose both.

l4u532:
> Thanks for the updating the code. I think we're very close now. I left some comments inline.

Incorporated the requested changes.

FYI, there is a phenomenon which I do not yet quite understand. I assume its harmless. When updating pitches, "id" is added - additionally to "_id":

```
    pitches: [
        {
            id: 'a011077d-58bd-4edd-9d9c-e5f3a4520add', # <-- duplicate, when updating the pitch
            parentId: UUID("97737f2ddac94ecda7645490ec1c40a0"),
            pitchNumber: 1,
            grades: {
                ewbank: '19'
            },
            type: {
                sport: false,
                alpine: true
            },
            length: 20,
            boltsCount: 6,
            description: 'Updated first pitch description',
            _id: UUID("a011077d58bd4edd9d9ce5f3a4520add") # <-- original
        },
```

src/model/MutableClimbDataSource.ts
```diff
@@ -125,6 +125,23 @@ export default class MutableClimbDataSource extends ClimbDataSource {
...
+        ? pitches.map((pitch): IPitch => {
+          if (pitch.pitchNumber === undefined) {
+            throw new UserInputError('Each pitch in a multi-pitch climb must have a pitchNumber representing its sequence in the climb. Please ensure that every pitch is numbered.')
+          }
+
+          return {
+            ...pitch,
+            _id: muid.from(pitch.id ?? muid.v4()), // generate MUUID if not present
+            parentId: muid.from(pitch.parentId ?? newClimbIds[i]),
+            pitchNumber: pitch.pitchNumber
```
vnugent: @l4u532 the spread operator `...pitch` will add the **id** field in addition to _id.

vnugent (changes requested):

vnugent:
> FYI, there is a phenomenon which I do not yet quite understand. I assume its harmless. When updating pitches, "id" is added - additionally to "_id"

As [mentioned](https://github.com/OpenBeta/openbeta-graphql/pull/350#discussion_r1336086813) the spread operator `...pitch` in the db update code introduces the id field.

Once you fix that, we're good to merge.  I think we may need to create an additional create/update/delete pitch API to make it easier to work with pitch data.

l4u532:
> > FYI, there is a phenomenon which I do not yet quite understand. I assume its harmless. When updating pitches, "id" is added - additionally to "_id"
> 
> As [mentioned](https://github.com/OpenBeta/openbeta-graphql/pull/350#discussion_r1336086813) the spread operator `...pitch` in the db update code introduces the id field.
> 
> Once you fix that, we're good to merge. I think we may need to create an additional create/update/delete pitch API to make it easier to work with pitch data.

Sorry for the belated fix, I was on vacation. I am kindly asking you to review this hopefully final fix :) @vnugent

vnugent:
I'm traveling at the moment.  Will get back to you by next week.

vnugent (approved):

PR #361, opened 2023-11-13
refactor: remove .js ext from ALL local imports across the codebase

enapupe:
Ideally, imports should reference the source file, not the build file. 

the trigger for this change was a requirement for jest to work directly with typescript files rather than the build output. but not just that, this change should move the project to a more standard direction and also improve code editors to better understand the code structure since now they reference the source files.

While there, 
- I also added a trigger for github test action when pushing the branch `test`.
- Added yarn caching via `actions/setup-node@v3`


package.json
```diff
@@ -66,7 +66,8 @@
   "scripts": {
     "lint": "yarn ts-standard",
     "fix": "yarn ts-standard --fix",
-    "test": "yarn build && cross-env NODE_OPTIONS=\"--experimental-vm-modules\" jest --runInBand",
+    "test": "jest",
```
vnugent: `jest test': would this pass locally?  As the way climb and area CRUD tests are written, they need to run sequentially within their suite, hence the --runInBand flag.

enapupe: Not really, `runInBand` continues to be a base requirement for tests to pass. I've restored it.

vnugent (approved):

PR #364, opened 2023-11-14
minor improvements to DX: jest running on src files; tweaks to GH actions

enapupe:
- this PR changes jest to run on source files so building is no longer a necessary step, it also enables TDD (with --watch flag). 
- unify GH actions into a single file that only runs the docker push segment when the build is a tag release; the goal is to have manual deployments when needed and not when development is merged.

Hopefully I'm not breaking prod this time.


.github/workflows/nodejs.yml
```diff
@@ -4,32 +4,71 @@ on:
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Build docker image
+        uses: docker/build-push-action@v2
+        with:
+          context: ./
+          file: ./Dockerfile
+          builder: ${{ steps.buildx.outputs.name }}
+          push: true
+          tags: vnguyen/openbeta-graph-api:latest
```
vnugent: Should we use Git tag instead of latest?  Example: `vnguyen/openbeta-graph-api:v1.0.2`

enapupe:
I believe this is missing just that curl command related to ArgoCD.

PR #365, opened 2023-12-09
fix: expose `boltsCount` and `pitches` in GraphQL mutation SingleClimbChangeInput

l4u532:
The Mutation [SingleClimbChangeInput](https://github.com/l4u532/openbeta-graphql/blob/31e0188fe613dfea67ae18247bce1b1c3fc6499f/src/graphql/schema/ClimbEdit.gql#L31-L50) is actually missing the fields `boltsCount` and `pitches`. Hence, these cannot be safely updated using the GraphQL Mutation. This should finalise #350.

What’s changed:
- exposed `boltsCount` and `pitches` in Mutation `SingleClimbChangeInput`
- renamed `Pitch.grades` to `Pitch.grade` and harmonised behaviour of `Climb.grade` with `Pitch.grade`
- made pitches’ `parentId` optional, since individual pitches cannot be updated anyway without providing a *parent climb’s `uuid`* (== a pitch’s `parentId`); also, this was preventing comfortably adding a climb + its pitches in one, single Mutation
- added a UserInput check to verify that each `pitchNumber` is unique for a given Mutation


src/graphql/schema/ClimbEdit.gql
```diff
@@ -42,32 +42,45 @@ input SingleClimbChangeInput {
...
+input PitchInput {
+  id: ID!
+  parentId: ID!
+  pitchNumber: Int!
+  grades: GradeTypeInput
+  disciplines: DisciplineType
+}
+
 input DisciplineType {
-  "https://en.wikipedia.org/wiki/Traditional_climbing"
```
l4u532, line 49: Removed these comments to clean up code

vnugent (approved):

Issue #377, opened 2024-01-04
Create an export job to generate crag-only geojson


vnugent:
### Use case

As a consumer of OB data, when building interactive maps, I want to have a vector map layer with all the crags included (as points), similar to how rivers, roads, buildings, etc are parts of a layer.  This layer can be used by openbeta.io or anyone using mapbox-gl or maplibre.
- [x] create crag exporter job
- [x] calculate area polygons
- [x] manually import files to MB studio

vnugent:
the cronjob now exports **crag.geojson.gz** to https://github.com/OpenBeta/openbeta-export
I'll set it to run daily.  We'll need a separate task to upload it to Mapbox studio.

---
<img width="974" alt="Screenshot 2024-01-15 at 8 33 18 AM" src="https://github.com/OpenBeta/openbeta-graphql/assets/3805254/7451c7bf-f2ea-4d3e-b8ea-28819d4fa515">

strukturart:
where can I find the exports ?

PR #380, opened 2024-01-15
feat: push crag data files to openbeta-export repo

vnugent:
#377


vnugent:
I pushed this old branch by accident

PR #407, opened 2024-11-04
Structure Tools

CocoisBuggy:
I'm experimenting with a feature for area navigation and climbing data input and I don't want to hammer the current GQL queries to get the data out, especially since I mostly just need pointers indicating the overall structure,

I've got the speed down to an acceptable time, but really it **should** be faster. I need to eliminate the lookup stage in the pipeline, which means embedding an ObjectId for an area's parent - I just haven't done it yet.

Queries can't run over 900ms at the moment, which I feel is probably too lenient. I'm just putting these commits here for now while I work on the rest of the PR.

## Other unrelated commits

I can kill these from the PR. They are not crucial, they mostly relate to making the development environment a little more forgiving

1. I added shell.nix - I'm on nixos 
2. made replica set env var optional during seeding
3. added a config in the vscode settings to prevent vscode extensions from mangling existing shell scripts


CocoisBuggy:
I've had too much time away from this PR and would like to imagine that actually the issues herein can be resolved by  #428 which should improve the integrity of areas queries. I will want to revisit this at a later time but the code can be better written assuming #428 can get merged at some point

Issue #126, opened 2022-09-06
Create an "update parent" mutation


vnugent:
### Use case
I want to move an area to be under a different parent.

To do:
- [ ] Update parent id 
- [ ] Update path tokens (look at area renaming code for ways to recursively update children's paths)
- [ ] Update ancestors
- [ ] Unit test

CocoisBuggy:
I've been needing this for a while and have decided to tackle it in connection with another feature I am needing - which would benefit greatly from knowing how group structures are changed and validated.

I will create a PR soon, but the gist is that I am going to add a new field to the area collection by the name of 'parent'. This is the source of truth from which

 - path tokens
 - ancestors
 - children

will be derived (Not to mention my other structure hierarchy query will be able to run 2-4x faster through the mongo pipeline)

Issue #422, opened 2024-11-19
API v2


vnugent:
A placeholder issue to outline v2.  More TBD.
1.  Keep the existing API
2. Create a v2 API - this will allow our web and mobile app to incrementally adopt new queries/mutations

Chores:
- #429 
- #430 
- #432

Issue #352, opened 2023-08-25
Ancestor property contains incorrect data


ClimbFlyCamp:
Looking at the ancestor property on an area I see the following:

```
"ancestors":"1db1e8ba-a40e-587c-88a4-64f5ea814b8e,60bc2823-f54f-59fa-9443-f61a87765009,8293b197-46ad-5f05-8f6d-b72387f2f191"
```

These ids do not correlate to any ids inside of the data. Think these should be converted to a similar model that the children field follows:

```
"children":["63689d10e80bff5a994f15ad"]
```

The children field correctly points to `_id` of the sub area, while the `ancestors` field is incorrect. There is no way to currently match an ancestor to a sub area

musoke:
Could you clarify this issue? As far as I can tell, each of the three returned ids corresponds to an area that exists in the production database:
- https://openbeta.io/crag/1db1e8ba-a40e-587c-88a4-64f5ea814b8e
- https://openbeta.io/crag/60bc2823-f54f-59fa-9443-f61a87765009
- https://openbeta.io/crag/8293b197-46ad-5f05-8f6d-b72387f2f191

It is a bit odd that the three are in a single comma separated string.

ClimbFlyCamp:
> Could you clarify this issue? As far as I can tell, each of the three returned areas exists in the production database:
> 
>     * https://openbeta.io/crag/1db1e8ba-a40e-587c-88a4-64f5ea814b8e
> 
>     * https://openbeta.io/crag/60bc2823-f54f-59fa-9443-f61a87765009
> 
>     * https://openbeta.io/crag/8293b197-46ad-5f05-8f6d-b72387f2f191

If you look at the exported data, here is an example:

https://github.com/OpenBeta/openbeta-export/tree/production/usa/alabama/bucks_pocket

Searching on any of those ancestor ids brings up the following:

https://github.com/search?q=repo%3AOpenBeta%2Fopenbeta-export%201db1e8ba-a40e-587c-88a4-64f5ea814b8e&type=code

You can see there are not actual hits, besides in the ancestor field. These do not point to the _id field of an area like the children do. With this data there is no way for me to link an ancestor to a sub area

ClimbFlyCamp:
There was a discussion here:

https://discord.com/channels/815145484003967026/815490505143484427/1144641711209971756

musoke:
Ah, I see!  It's just a bug in exported data.

[...]

Issue #316, opened 2023-06-14
Slow query when updating a lot of path tokens


vnugent:
When working on PR #314 my laptop CPUs peaked for about 1 minute when updating a large tree.

Steps to reproduce:
1.  Rename a US state
2. Watch mongo log (`docker logs -f <mongo container id>`)

While the update will take around 2-3s, I think the change history code causes CPUs to peak and fan running at high speed for about 1 minute.

https://github.com/OpenBeta/openbeta-graphql/blob/9e3318ee3a3e1b8fb61e3500c4e39b2c5b6d5985/src/model/ChangeLogDataSource.ts#L36

mongo complains about Slow query

```javascript

{"t":{"$date":"2023-06-14T04:55:16.739+00:00"},"s":"I",  "c":"COMMAND",  "id":51803,   "ctx":"conn1744","msg":"Slow query","attr":{"type":"command","ns":"openbeta.$cmd","command":{"update":"change_logs","updates":[{"q":{"_id":{"$oid":"64894824fd36d9b6032697e2"}},"u":{"$setOnInsert":{"createdAt":{"$date":"2023-06-14T04:55:05.320Z"}},"$push":{"changes":{"$each":[{"_id":{"_data":"8264894828000000012B022C01002C01EC6E5A1004A1760F9BEFF84607B0673B9FB1CC630946645F696400646467AA00E622317BDB3C18AE0004"},"dbOp":"update","fullDocument":{"_id":{"$oid":"6467aa00e622317bdb3c18ae"},"area_name":"Átya tpísh (North face)","shortCode":"","climbs":[{"$uuid":"5db94c91-4482-529a-a113-30c2d969a9d6"},{"$uuid":"34f74ebd-3d80-5a5e-b571-532d173526a6"},{"$uuid":"5ecc5de6-c6d9-57c6-870a-348bc80f14ee"},

...
```

PR #428, opened 2024-11-28
Coco/area relational integrity

CocoisBuggy:
Fulfils the feature request in #126 

There's a bit happening here, since I made some performance driven changes to help me with a different PR (#407) I tried to knock out the major test cases but the nature of this particular feature is a little thorny because of how we utilize de-normalized data.

The commit history is a little messy and I am frankly feeling a bit lazy, which is why I have not squashed them. Been moving between machines a bunch and reworked the approach a few times to get speed tuned in.
If someone would like me to squash then I will try and motivate myself to do it.

The key takeaway in terms of changes that will affect other devs is that a new field has been introduced on the area document from which all other ancestry / children are derived. the `parent` field points to the areas direct parent. The pathTokens and ancestors fields have been colocated into a single array of sub documents so that the update propogations can happen all together - not a big deal now, but gives us a sane path to embed  more de-normalized data if we need to.

@vnugent I'd like your particular eyes for these changes.


vnugent:
@CocoisBuggy can you please explain what problems this PR is trying to solve?  Maybe a diagram or example would help.  Thanks!
Not sure if you notice I created a rough plan for v2 #422 where we can address some of the major issues.  Perhaps we can discuss larger impact changes in v2?

CocoisBuggy:
Sure thing, sorry I suppose I wasn't super clear in my explanation.


The tldr is:
- performance
- code sanity
- collate demoralized fields together
- allow updating of area parents to reorganize areas
- deconvolute data structure

## Who should be affected?

Only developers working on the back end. There is a scary number of changes, but I haven't touched the way in which the data is accessed by the api so there **shouldn't** be a problem - but now that I'm writing that it may be just as well for me to check that the existing tests cover the behavior.

Developers shouldn't have to change their mental model too much, since all the old data Is still available - and is recomputed when it's dependencies update.

## Performance

In my other pr I'm working on some queries that I need for a feature I'm interested in, but performing the lookup for parents of nodes was always a substantial hit when I was looking at my profiler. Partly this is a consequence of storing the ancestors as a comma delimited string which seemed to always be a performance sink whenever I touched it. When using the children array I found a couple of strange hydras when performing deep queries.

After struggling to get the queries down below 100-200ms I decided to look elsewhere for a solution.

## Code Sanity

While I was looking into area update and coming to terms with the existing denormalization strategies I found myself yearning for a single approach to be used so that it could all be debugged and understood together - especially since they're absolutely related code work.

It also made me increasingly perplexed that the children were specified as an array of ObjectId but the ancestors is a string - and a string of uuid values at that.

My solution to future-proof against this problem getting worse over time and also resolve the inconsistencies I decided to put the related data together.

## Area Structure Updating

When I went to implement this feature neither the array of children nor the ancestor string felt like good ways to express which area had which parent. I decided to make these explicitly derived fields, and then simply put a parent field on each area. The difference here is slightly more than philosophical, but without going into too much detail my belief is that this is the easiest way to not get stuck in side effect hell. I can expand more on that if needed, but for the time being it suffices to say that there's no way to accidentally end up with areas belonging to multiple parents, which would have cascading bugs.

So now you need only update an areas parent, and then recompute the paths from the altered node (which is quite speedy).

## conclusion

The pr is more or less worthless if we want to specifically avoid a parent field pattern. It should work well with the v2 milestone, especially for phasing out uuid under the hood. I can do more legwork to make sure it doesn't change the existing api, as it shouldn't be necessary for this pr

vnugent:
@CocoisBuggy Thanks for the large PR.  Can we discuss this after the new year?

CocoisBuggy:
For sure, merry Christmas @vnugent !

CocoisBuggy:
Happy new year fellow climbers! I hope everyone had a good christmas. Down here in SA we are so far out of season we do a nightly prayer for our poor skin.

I thought I'd resolve the merge issues as a nice little start to the year. I'm going to copy the contents of my previous commit as it contains a question I now have:

```
My confidence in this deletion is a little lower than it initially was.
I remember checking the code archeology the first time around and
thinking that we must have been done with importing data (especially
since descriptions have become mutable and I had assumed importing
might result in unintensional overwrites.)

However this change to the code indicates otherwise. Could someone
indicate where in the workflow this code is still in use? If it's still
in use then obviously I actually do need to fix this up for the PR to
make any sense...
```

CocoisBuggy:
#316 
#352
#126
#458
are all partially or fully resolved in this pr, any eyes on this or feedback would be greatly appreciated

Issue #420, opened 2024-11-12
Could we use Vitest over Jest?


CocoisBuggy:
My suspicion is that vitest would let us clean up a lot of the test infrastructure using their [fixtures](https://vitest.dev/guide/test-context). I have strange issues with jest that I consistently cannot explain and it is definitely verrrry slow. I'll attach a PR to this issue so that some numbers can be put alongside this musing.

vnugent:
I'm not familiar with Vite and also don't have a strong preference.  Do we have to switch to vite for build as well?  Can it be adopted incrementally?

CocoisBuggy:
My experience with vitest has always been that it's a drop-in replacement for jest. Config is a little different (it uses the vite config style), but api is otherwise identical to jest/node test/bun test.

Vite is not required to use vitest. Vitest uses vite under the hood and is part of that same ecosystem.

I have a pretty specific idea in connection with adopting vitest, and my hope is that I can speed up the tests a lot. I'll create a PR for it at some point to explain my idea - I like to run tests as I code so speed is a big priority when I build test suites

Issue #318, opened 2023-06-15
Create a test harness for writing unit tests against GQL layer


vnugent:
Write some abstraction / a test harness to make it easier to write unit tests against the GQL layer.

PR #437, opened 2024-12-14
A vitest example

CocoisBuggy:
All but one test passing, Just thought I would share the outcome so my issue doesn't seem so arbitrary:

## Headlines

 - Jest single band tests : **77s - 90s**
 - Vitest Threaded pool: **7s - 11s** _(37.68 seconds of execution time, if you sum all test runner execution times together, but since they run together this is not what users experience)_
 - Vitest Forked pool: **12s - 14s** Straight downgrade from threads but still quick.

It's a roughly 10x improvement which is frankly not what I had hoped to achieve, the tests are roughly as fast as I'd like them to be when running in a `--watch` context

## Objectives

tests should run faster, and `--watch` should be a little more usable to tighten up development feedback loops. Ideally, I'd also like the test runners to have **much** less noise, since we use a LOT of code to scaffold test state that is not actually a part of what particular state is under examination in the test. This slows comprehension and makes the test files very bulky and cumbersome.

## Remaining issues

Obviously, one of the tests is still not passing so... that's not good. It's a history test, and as soon as you have forked control flows parallel testing gets a little tricky. The reason I stopped debugging the test was because I could not properly rationalize what the intended behavior was **supposed to be**. I need to take a longer think about the changelog system we have, but I thought I'd leave this here to explain #420 a little bit - we do not need to pursue this PR, it is not a priority at the moment


CocoisBuggy:
I've merged all new needed stuff into this branch in case anyone wants to take a peak at a more up-to-date proposal

Vitest runs in *~8s*
Existing test strat: *~90s*
Speedup: **1125%** (11.25x)  :rocket: 

It should be said that this is not necessarily all down to vitest being that much faster than jest (though in my experience it is a bit faster). The real benefit comes from isolating the test states properly using vitest's fixture api. This lets us run the tests in parallel using their little rust runners under the hood.

I will also note that two of the history tests fail because I do not understand the history behavior well enough to redesign the tests

CocoisBuggy:
I need a little help with understanding how our history is supposed to work so I can write strict tests for it, but other than that this testing suite is a straight upgrade :small_airplane:

Hopefully this pleases someone as much as it pleases me :sweat_smile:

CocoisBuggy:
Another checkpoint: All tests are ported.

This means this PR now represents a drop-in replacement for the current tests, except it will run much faster and the test infrastructure is designed with the full application in mind

A good day to all :blush:

CocoisBuggy:
I really want to lobby for this since it represents a huge speed improvement in the feedback loop for developing features and fixing bugs. Does anyone have any concerns?

CocoisBuggy:
#318 would get addressed by this. All tests are comprehensively instrumented with fixtures that make writing consistent tests easier

PR #439, opened 2024-12-22
Will kencel/tick date handling

wkencel:
The error occurred because some ticks had malformed dates that couldn't be parsed into valid Date objects, causing the getTime() function to fail. This change gracefully handles those cases while maintaining the intended noon UTC standardization.

This change:

- Handles both Date objects and string dates (references test data in src/model/__tests__/ticks.ts lines 19 and 31)

- Validates dates properly before using getTime() (fixes the error in the GraphQL response)

- Sets invalid dates to null instead of throwing errors (maintains data consistency)

- Maintains compatibility with the America/Denver timezone standardization from the migration (references db-migrations/0003-date-climbed-to-date.js)

- Matches the test expectations for date handling (references src/__tests__/ticks.ts line 193)

I also added a test. screenshots on repro of error, successful testing after change and successful running test

Repro1
repro2

After Change
VerificationOfFix
Screenshot 2024-12-21 at 4 50 59 PM

### DIFF
--- a/src/__tests__/ticks.ts
+++ b/src/__tests__/ticks.ts
@@ -204,7 +204,7 @@ describe('ticks API', () => {
               name: 'Updated Route One',
               climbId: 'new climb id',
               userId: userUuid,
-              dateClimbed: '2022-11-10',
+              dateClimbed: new Date('2022-11-10T12:00:00Z'),
               grade: 'new grade',
               source: 'OB'
             }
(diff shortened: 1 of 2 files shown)


vnugent:
@all-contributors @wkencel for code & ideas

vnugent:
@all-contributors add @wkencel for code and ideas

Issue #419, opened 2024-11-12
Ticks mutation should accept nullable style and attempt type


Vichy97:
Style and attempt type are marked nullable in the schema, but failing to provide them will return an error. We should accept null. It doesn't make sense to pass "N/A" when null would suffice

glassbead0:
This is handled in https://github.com/OpenBeta/openbeta-graphql/pull/444

Issue #411, opened 2024-11-10
Structured data for `attemptType`


CocoisBuggy:
There has been some confusion about what data *exactly* to put in this field, and it would benefit us in the long-run to constrain this field for data aggregation purposes.

 - [ ] Add an enumeration with reasonable attemptTypes
 - [ ] Allow for some array of these attemptTypes to be assigned to a given tick - There is some philosophizing to be done on this point, but realistically you can have some of these types happen in conjunction.
- [ ] Allow fetching of the enum from the database - This will be a changing set, so we shouldn't hard-code values accross the various services.

Vichy97:
This should also apply to tick styles. Both should be an enum

julianlam:
For what it's worth, a tick style doesn't seem to be relevant in a bouldering context. The field should be hidden in front-end, and if not already done so, the backend should allow `null` as an acceptable value (related; #419)

There are additional restrictions one can add (e.g. does the submitted style and attempt type combination make sense?), so things like `Boulder/Pinkpoint` are not allowed, but I think that might be overkill.

vnugent:
Copying @Vichy97 comment from Discord

**style**: only applicable to roped climbs.
Possible values:
- solo
- tr
- follow
- lead

**attempt**: Only applicable to boulders or roped climbs with "lead" style
Possible values for roped climbs
- attempt
- redpoint
- pinkpoint
- flash
- onsight

Possible values for boulders
- attempt
- send
- flash

anything that does not fall into these categories should throw an error

--

- Question: For roped climbs is **attempt** the same as hangdogging?
- Todo:  we need to support **style** field

Vichy97:
Ya I think attempt can represent Hung/Fell. The client can display this differently than it is recorded in the backend. Also we do have a style field, it just doesn't get imported from MP correctly

vnugent:
For **lead style** should we make it clear whether it's a free attempt or it's aid/french-free

prith-k:
Hi, is anyone actively working on this? Just want to make sure I'm not duplicating work. I'm new here and this seems like a reasonable "good first issue".

CocoisBuggy:
@PrithipalKhalsa I don't think anyone is working on this right now, it may well be a reasonable first issue. Let me know if you need a hand with getting up and running

PR #444, opened 2025-01-22
feat: adding tick restriction for attempt type

prith-k:
Addressing #411:
Restricting attemptType on bouldering to : Send, Flash or Attempt
Only Routes can be a style of: Lead, Solo, TR, Follow (aka no style for bouldering)
Solo, TR, and Follow style cannot be: Attempt, Redpoint, Pinkpoint, Flash, or Onsight

Addressing: #419 
@Vichy97 I simply made this a non-required field. Which is how I interpreted this issue.

I also had to do some larger updates to the tick test files. Since adding and updating a tick requires verification of the climb style.

I am newer to graphql so if I've missed anything or broke convention please let me know :)

### DIFF
--- a/src/db/TickSchema.ts
+++ b/src/db/TickSchema.ts
@@ -16,10 +16,10 @@ export const TickSchema = new Schema<TickType>({
   notes: { type: Schema.Types.String, required: false },
   climbId: { type: Schema.Types.String, required: true, index: true },
   userId: { type: Schema.Types.String, required: true, index: true },
-  style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow', 'Aid'], required: false },
-  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Redpoint', 'Send', 'Attempt'], required: false, index: true },
+  style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow', 'Aid', 'Boulder', null], required: false },
+  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Redpoint', 'Send', 'Attempt', null], required: false, index: true },
   dateClimbed: { type: Schema.Types.Date },
-  grade: { type: Schema.Types.String, required: true, index: true },
+  grade: { type: Schema.Types.String, required: false, index: true },
   // Bear in mind that these enum types must be kept in sync with the TickSource enum
   source: { type: Schema.Types.String, enum: ['MP', 'OB'] as TickSource[], required: true, index: true }
 })
--- a/src/db/TickSchema.ts
+++ b/src/db/TickSchema.ts
@@ -16,8 +16,8 @@ export const TickSchema = new Schema<TickType>({
   notes: { type: Schema.Types.String, required: false },
   climbId: { type: Schema.Types.String, required: true, index: true },
   userId: { type: Schema.Types.String, required: true, index: true },
-  style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow'], required: false },
-  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Redpoint', 'Pinkpoint', 'Frenchfree', 'Fell/Hung', 'Send', 'Attempt'], required: false, index: true },
+  style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow', 'Aid'], required: false },
+  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Send', 'Attempt'], required: false, index: true },
   dateClimbed: { type: Schema.Types.Date },
   grade: { type: Schema.Types.String, required: true, index: true },
   // Bear in mind that these enum types must be kept in sync with the TickSource enum
--- a/src/db/TickSchema.ts
+++ b/src/db/TickSchema.ts
@@ -16,8 +16,8 @@ export const TickSchema = new Schema<TickType>({
   notes: { type: Schema.Types.String, required: false },
   climbId: { type: Schema.Types.String, required: true, index: true },
   userId: { type: Schema.Types.String, required: true, index: true },
-  style: { type: Schema.Types.String, required: true, default: '' },
-  attemptType: { type: Schema.Types.String, required: true, index: true, default: '' },
+  style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow'], required: false },
+  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Redpoint', 'Pinkpoint', 'Frenchfree', 'Fell/Hung', 'Send', 'Attempt'], required: false, index: true },
   dateClimbed: { type: Schema.Types.Date },
   grade: { type: Schema.Types.String, required: true, index: true },
   // Bear in mind that these enum types must be kept in sync with the TickSource enum
--- a/src/db/TickSchema.ts
+++ b/src/db/TickSchema.ts
@@ -16,8 +16,8 @@ export const TickSchema = new Schema<TickType>({
   notes: { type: Schema.Types.String, required: false },
   climbId: { type: Schema.Types.String, required: true, index: true },
   userId: { type: Schema.Types.String, required: true, index: true },
-  style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow', 'Aid', 'Boulder', null], required: false },
-  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Redpoint', 'Send', 'Attempt', null], required: false, index: true },
+  style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow', 'Aid', 'Boulder'], required: false },
+  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Redpoint', 'Send', 'Attempt'], required: false, index: true },
   dateClimbed: { type: Schema.Types.Date },
   grade: { type: Schema.Types.String, required: false, index: true },
   // Bear in mind that these enum types must be kept in sync with the TickSource enum
--- a/src/db/TickSchema.ts
+++ b/src/db/TickSchema.ts
@@ -17,7 +17,7 @@ export const TickSchema = new Schema<TickType>({
   climbId: { type: Schema.Types.String, required: true, index: true },
   userId: { type: Schema.Types.String, required: true, index: true },
   style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow', 'Aid'], required: false },
-  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Send', 'Attempt'], required: false, index: true },
+  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Redpoint', 'Send', 'Attempt'], required: false, index: true },
   dateClimbed: { type: Schema.Types.Date },
   grade: { type: Schema.Types.String, required: true, index: true },
   // Bear in mind that these enum types must be kept in sync with the TickSource enum
--- a/src/db/TickTypes.ts
+++ b/src/db/TickTypes.ts
@@ -16,11 +16,11 @@ export type TickSource =
    */
   'MP'
 
-export type TickStyle = 'Lead' | 'Solo' | 'TR' | 'Follow' | 'Aid'
+export type TickStyle = 'Lead' | 'Solo' | 'TR' | 'Follow' | 'Aid' | 'Boulder'
 export type TickAttemptType = 'Onsight' | 'Flash' | 'Pinkpoint' | 'Frenchfree' | 'Send' | 'Attempt' | 'Redpoint'
 
 export const TickSourceValues: TickSource[] = ['OB', 'MP']
-export const TickStyleValues: TickStyle[] = ['Lead', 'Solo', 'TR', 'Follow', 'Aid']
+export const TickStyleValues: TickStyle[] = ['Lead', 'Solo', 'TR', 'Follow', 'Aid', 'Boulder']
 export const TickAttemptTypeValues: TickAttemptType[] = ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Send', 'Attempt', 'Redpoint']
 
 /** Ticks
@@ -74,8 +74,8 @@ export interface TickInput {
   style?: TickStyle
 
   /**
-   * Describe the type of successful attempt that was made here.
-   * Attempt, Flash, Redpoint, Onsight, would be examples of values you might find here.
+   * Describe the type of attempt that was made here.
+   * Attempt, Flash, Redpoint, Onsight, Attempt would be examples of values you might find here.
    * This is again a free-form field. Data of practically any descriptive nature may find
    * itself here.
    */
@@ -94,7 +94,7 @@ export interface TickInput {
    * proper operation when importing ticks for entities that cannot be located
    * within OpenBeta's database.
    * */
-  grade: string
+  grade?: string
 
   /**
    * we support any number of sources for tick import or native ticks,
--- a/src/db/TickTypes.ts
+++ b/src/db/TickTypes.ts
@@ -16,7 +16,12 @@ export type TickSource =
    */
   'MP'
 
+export type TickStyle = 'Lead' | 'Solo' | 'TR' | 'Follow'
+export type TickAttemptType = 'Onsight' | 'Flash' | 'Redpoint' | 'Pinkpoint' | 'Frenchfree' | 'Fell/Hung' | 'Send' | 'Attempt'
+
 export const TickSourceValues: TickSource[] = ['OB', 'MP']
+export const TickStyleValues: TickStyle[] = ['Lead', 'Solo', 'TR', 'Follow']
+export const TickAttemptTypeValues: TickAttemptType[] = ['Onsight', 'Flash', 'Redpoint', 'Pinkpoint', 'Frenchfree', 'Fell/Hung', 'Send', 'Attempt']
 
 /** Ticks
  * Ticks represent log entries for a user's climbing activity. They contain
@@ -66,7 +71,7 @@ export interface TickInput {
    * If this is a native tick, you can enforce updated values here by referencing
    * the climb document (climbId -> climbs:uuid)
    */
-  style: string
+  style: TickStyle
 
   /**
    * Describe the type of successful attempt that was made here.
@@ -74,7 +79,7 @@ export interface TickInput {
    * This is again a free-form field. Data of practically any descriptive nature may find
    * itself here.
    */
-  attemptType: string
+  attemptType: TickAttemptType
 
   /**
    * Not the same as date created. Ticks can be back-filled by the user, and do
--- a/src/db/TickTypes.ts
+++ b/src/db/TickTypes.ts
@@ -17,11 +17,11 @@ export type TickSource =
   'MP'
 
 export type TickStyle = 'Lead' | 'Solo' | 'TR' | 'Follow' | 'Aid'
-export type TickAttemptType = 'Onsight' | 'Flash' | 'Pinkpoint' | 'Frenchfree' | 'Send' | 'Attempt'
+export type TickAttemptType = 'Onsight' | 'Flash' | 'Pinkpoint' | 'Frenchfree' | 'Send' | 'Attempt' | 'Redpoint'
 
 export const TickSourceValues: TickSource[] = ['OB', 'MP']
 export const TickStyleValues: TickStyle[] = ['Lead', 'Solo', 'TR', 'Follow', 'Aid']
-export const TickAttemptTypeValues: TickAttemptType[] = ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Send', 'Attempt']
+export const TickAttemptTypeValues: TickAttemptType[] = ['Onsight', 'Flash', 'Pinkpoint', 'Frenchfree', 'Send', 'Attempt', 'Redpoint']
 
 /** Ticks
  * Ticks represent log entries for a user's climbing activity. They contain
--- a/src/db/TickTypes.ts
+++ b/src/db/TickTypes.ts
@@ -16,12 +16,12 @@ export type TickSource =
    */
   'MP'
 
-export type TickStyle = 'Lead' | 'Solo' | 'TR' | 'Follow'
-export type TickAttemptType = 'Onsight' | 'Flash' | 'Redpoint' | 'Pinkpoint' | 'Frenchfree' | 'Fell/Hung' | 'Send' | 'Attempt'
+export type TickStyle = 'Lead' | 'Solo' | 'TR' | 'Follow' | 'Aid'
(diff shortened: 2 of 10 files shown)


Vichy97:
This looks good to me from a functionality standpoint (I'll defer to others for the code review since I'm not a web developer). I think right now this can't merge until we fix MP imports though, otherwise import will always fail. I also wonder if we need a DB migration for existing ticks?

src/graphql/schema/Tick.gql
```diff
@@ -120,6 +120,40 @@ enum TickSource {
...
+enum TickStyle {
+  "Lead"
+  Lead
+  "Solo"
+  Solo
+  "tr"
+  TR
+  "Follow"
+  Follow
+}
```
glassbead0, line 154: Maybe we add 'Boulder'? It's a little counter-intuitive to me that this field is blank for boulders, IMHO setting the style to `Boulder` makes sense.

glassbead0:
@Vichy97 @PrithipalKhalsa 

I'm hoping over here from https://github.com/OpenBeta/open-tacos/pull/1259, which fixes MP imports and relates to this PR. I have a couple thoughts and would be happy to help out on this:


1) I'm going to update my PR (1259) to not use 'N/A' for style or attemptType, and just use null instead. 
 
 2) Regarding a migration to fix the incorrect attemptyType for all previously imported MP ticks: I don't see how to do that  because whatever 'Lead Style' value came in from MP at time of import is gone (AFAICT). so right now, unless individual users have manually edited their ticks, the `attemptStyle` field for all MP imports is meaningless. Perhaps a migration to just null out all `attemptType`s of `Send` that have a `source` of `MP`?

3) I have a strong opinion, weakly held (happy to shut up about this if nobody agrees :) ), that both MP and OB schema/naming for ticks aren't great. The 2 fields (style and attemptType), aren't consistently used between routes/boulders, and things are not named super intuitively. eg: one of the possible values for `attemptType` is `Attempt`. I know what that means, but it takes some cognitive overhead to understand it, I also wonder if we should set `style` to `Boulder` for boulders, instead of leaving it blank. I left some comments in the code in this PR with my thoughts.

src/db/TickSchema.ts
```diff
@@ -16,8 +16,8 @@ export const TickSchema = new Schema<TickType>({
   notes: { type: Schema.Types.String, required: false },
   climbId: { type: Schema.Types.String, required: true, index: true },
   userId: { type: Schema.Types.String, required: true, index: true },
-  style: { type: Schema.Types.String, required: true, default: '' },
-  attemptType: { type: Schema.Types.String, required: true, index: true, default: '' },
+  style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow'], required: false },
+  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Redpoint', 'Pinkpoint', 'Frenchfree', 'Fell/Hung', 'Send', 'Attempt'], required: false, index: true },
```
glassbead0: I'm not the biggest fan of the field `attemptType` having a possible value of `Attempt`. Perhaps something like `Fell` (for boulders), instead? We use `Fell/Hung` for routes, so I feel like `Fell` makes sense for boulders.

glassbead0: Also, this is perhaps an uncommon use case, but are there enough people who would want an `Aid` option? I like having `FrenchFree` as an option, and `Aid` can certainly be considered distinct from FrenchFree.

src/graphql/schema/Tick.gql
```diff
@@ -120,6 +120,40 @@ enum TickSource {
...
+  "Onsight: Sending a route fist try with no prior knowledge"
+  Onsight
+  "Flash: Sending a route first try with prior knowledge"
+  Flash
+  "Redpoint: Sending a route with no falls"
+  Redpoint
+  "Pinkpoint: Sending a route with no falls on pre-placed gear"
+  Pinkpoint
+  "French Free: Doing the route with pulling on gear or draws"
+  French_Free
```
glassbead0: In other places, you're using `Frenchfree`, but here's its `French_Free`. is that intentional? (I'm new to graphql, so maybe this is correct in this context)

src/graphql/schema/Tick.gql
```diff
@@ -120,6 +120,40 @@ enum TickSource {
...
+  "Fell/Hung: Falling or hanging on gear"
+  Fell_Hung
```
glassbead0: same as above `Fell/Hung` vs `Fell_Hung`

src/graphql/schema/Tick.gql
Vichy97: I don't know if I agree with that. The UI (at least on the mobile app) just won't show this field. So for boulders this field doesn't make sense at all. This is similar to how MP shows ticks for boulders

src/db/TickSchema.ts
Vichy97: I'm not a boulderer so idk if fell or attempt makes more sense here. I would think some people might like attempt as opposed to fell if they got on a project just to work a particular move or sequence, but I'll defer to people who boulder more than I do.

Vichy97: > but are there enough people who would want an Aid option

I could go either way on this. The route grade is already marked with an aid rating. Again I don't aid climb ever and I don't know if I would ever use "french free", but I am open to including it or not including it

Vichy97:
> 2. Regarding a migration to fix the incorrect attemptyType for all previously imported MP ticks: I don't see how to do that  because whatever 'Lead Style' value came in from MP at time of import is gone (AFAICT). so right now, unless individual users have manually edited their ticks, the `attemptStyle` field for all MP imports is meaningless. Perhaps a migration to just null out all `attemptType`s of `Send` that have a `source` of `MP`?

I don't think we can migrate attemptStyle, but the attemptType fields that have already been migrated will not match what is required. New imports also won't match and will fail to import. That's my main concern WRT MP imports. Not matching fields also applies to existing ticks from OpenBeta

src/db/TickSchema.ts
glassbead0: I know `attempt` is pretty standard terminology, and MP certainly helped normalize it by using that term. I suspect people might have strong opinions about changing it, so I don't want to rock the boat.

Vichy97: I'm certainly not opposed to improving on what MP does, but I appreciate the simplicity of MP as opposed to something like 27 crags where theres like 20 options. I also know the majority of users probably will also be users of MP right now. I would lean towards mirroring MP for now, but I don't have a hard stance either way

src/model/TickDataSource.ts
```diff
@@ -57,10 +60,32 @@ export default class TickDataSource extends MongoDataSource<TickType> {
...
+      if ((!['Send', 'Flash', 'Attempt'].includes(tick.attemptType)) || (['Lead', 'Solo', 'Tr', 'Follow'].includes(tick.style))) {
+        throw new Error('Invalid attempt type or style for bouldering')
+      }
+    // Only applicable to boulders or roped climbs with "lead" style
+    } else if (!(tick.style === 'Lead')) {
+      if ((['Attempt', 'Redpoint', 'Pinkpoint', 'Flash', 'Onsight'].includes(tick.attemptType))) {
+        throw new Error('Invalid attempt type for a non-lead style')
+      }
+    }
+  }
```
glassbead0: This really gets in the weeds pretty fast, but since we have the option of `Fell/Hung` for Lead climbs, isn't that redundant with `Attempt`? 

On MP, `Attempt` can only be marked for boulders, and `Fell/Hung` can only be marked for roped climbs with 'Lead' style.

Vichy97: Actually I think we shouldn't support Fell/Hung. Keep in mind, these are not necessarily the user-facing values. I think attempt should map to Fell/Hung for climbs on the user-facing side, but stored as attempt in the server.

src/db/TickTypes.ts
```diff
@@ -16,7 +16,12 @@ export type TickSource =
    */
   'MP'
 
+export type TickStyle = 'Lead' | 'Solo' | 'TR' | 'Follow'
+export type TickAttemptType = 'Onsight' | 'Flash' | 'Redpoint' | 'Pinkpoint' | 'Frenchfree' | 'Fell/Hung' | 'Send' | 'Attempt'
```
Vichy97: No need for Fell/Hung here. That can be the user-facing version of "Attempt" IMO

glassbead0: Fair.

Also If these don't need to be user facing, then i could argue we get rid of `Send` as well, and use `Redpoint`. For boulders a `Redpoint` would get rendered on the front end as a `Send`, since it's standard nomenlature not to use redpoint in bouldering (even though conceptually it's basically the same).  And of course still leave 'Onsight' available for routes, and 'Flash' available for both routes and boulders.

Man.... I can see how 27 crags ends up with so many options... there actually are a lot of philosophical/linguistic/cultural issues at play here... lol.

All in all... whatever is done, I advocate for well documented code that makes it clear what means what.

glassbead0:
Is there a separate issue for the corresponding front-end change? Looks like at least https://github.com/OpenBeta/open-tacos/blob/develop/src/components/users/TickForm.tsx#L36 needs to be updated to match whatever is done here.

src/graphql/schema/Tick.gql
prith-k: That is not intentional, I'll fix that

src/db/TickTypes.ts
prith-k: I agree, although due to redpoint and send being used interchangeably with routes, I would suggest storing them both as send, and disregarding redpoint. Also since both dws, aid and tr would use 'send'.

prith-k:
I was mostly trying to interpret #411 and I was mostly trying to reduce conflict from MP imports.
Looking at the thread here, I'll suggest this:

Type: trad, sport, alpine, ice, mixed, aid
possible Styles: 'Lead', 'Solo', 'TR', 'Follow', 'Aid'
possible attempt types for Lead:
'Onsight', 'Flash', 'Redpoint', 'Pinkpoint', 'Frenchfree', 'Send', 'Attempt'
possible attempt types for Solo, TR and Follow: 'Send', 'Flash', 'Attempt', 'Onsight'

Type: DWS, Bouldering
possible styles: no field
possible attempt types: 'Send', 'Flash', 'Attempt', 'Onsight'

Type: tr
possible Styles: TR
possible attempt types:  'Send', 'Flash', 'Attempt', 'Onsight'

Type: aid
possible Styles: aid, follow
possible attempt types: send attempt (or possibly leave this blank)

src/db/TickTypes.ts
Vichy97: Ya I agree. Both should be Send

glassbead0: I'm curious why you guys think 'send' is better to store in the db instead of 'redpoint'. I think send is generally an ambiguous term, and Onsight, Redpoint and Flash are all specific instances of a 'send' (like a hierarchy). Even in bouldering I might casually say I sent, regardless of whether I got it 1st go (flash), or 5th go (technically a redpoint). DWS and TR can also meaningfully distinguish between OS, RP and Flash. (Aid is a different story perhaps... ugh)

I do agree things should match commonly used terminology on the front end (pretty much matching MP).

In pursuit of simplicity on the backend though, I'll make one last case to not use send:
* The 4 core attemptTypes of `Onsight, Redpoint, Flash, and Attempt` are applicable to almost everything with only one exception (no OS for boulders).
* `Frenchfree & Pinkpoint` are slightly special cases, but straightforward to deal with.

I don't want to be the annoying opinionated guy (especially as a new contributor), but I do know that a lot of climbers care about tick data. I'll shut up about it after this comment though and I'll happily review the code 😄  Also happy to chat or pair. (I'm not actually that strongly opinionated, but I do nerd out about linguistics and databases 🤓 )

src/model/TickDataSource.ts
```diff
@@ -57,10 +60,32 @@ export default class TickDataSource extends MongoDataSource<TickType> {
    * @returns the new/updated tick
    */
   async editTick (filter: TickEditFilterType, updatedTick: TickInput): Promise<TickType | null> {
+    await this.validateTick(updatedTick)
     const rs = await this.tickModel.findOneAndUpdate(filter, updatedTick, { new: true })
     return await rs?.toObject() ?? null
   }
 
+  private async validateTick (tick: TickInput): Promise<void> {
```
mikeschen: Just a request, could we make code a bit more readable, i.e. :
```
private async validateTick(tick: TickInput): Promise<void> {
  const climbIdAsUUID = muuid.from(tick.climbId);
  const climb = await this.climbModel
    .findOne({ _id: climbIdAsUUID, _deleting: { $eq: null } })
    .lean();

  if (!climb) {
    throw new Error('Climb not found');
  }

  // Check if the climb is exclusively a boulder
  const isBoulderingOnly = 
    climb.type.bouldering === true &&
    Object.keys(climb.type).every(
      key => key === 'bouldering' || climb.type[key] === false
    );

  if (isBoulderingOnly) {
    // Validate bouldering-specific attempt type and style
    const validBoulderingAttempts = ['Send', 'Flash', 'Attempt'];
    const invalidStyles = ['Lead', 'Solo', 'Tr', 'Follow'];

    if (
      !validBoulderingAttempts.includes(tick.attemptType) || 
      invalidStyles.includes(tick.style)
    ) {
      throw new Error('Invalid attempt type or style for bouldering');
    }
  } else {
    // Validate non-bouldering styles and attempt types
    const validRopedAttempts = ['Attempt', 'Redpoint', 'Pinkpoint', 'Flash', 'Onsight'];

    if (tick.style !== 'Lead' && validRopedAttempts.includes(tick.attemptType)) {
      throw new Error('Invalid attempt type for a non-lead style');
    }
  }
}
```

mikeschen (commented):
Just some code cleanup :)

Vichy97:
> I was mostly trying to interpret #411 and I was mostly trying to reduce conflict from MP imports. Looking at the thread here, I'll suggest this:
> 
> Type: trad, sport, alpine, ice, mixed, aid possible Styles: 'Lead', 'Solo', 'TR', 'Follow', 'Aid' possible attempt types for Lead: 'Onsight', 'Flash', 'Redpoint', 'Pinkpoint', 'Frenchfree', 'Send', 'Attempt' possible attempt types for Solo, TR and Follow: 'Send', 'Flash', 'Attempt', 'Onsight'
> 
> Type: DWS, Bouldering possible styles: no field possible attempt types: 'Send', 'Flash', 'Attempt', 'Onsight'
> 
> Type: tr possible Styles: TR possible attempt types: 'Send', 'Flash', 'Attempt', 'Onsight'
> 
> Type: aid possible Styles: aid, follow possible attempt types: send attempt (or possibly leave this blank)

I think this is ok, but I don't plan to display a picker for style for TR on the mobile client. This matches MP, but I also feel most people don't count TR as a "flash" or "onsight" or even really a "send"

src/db/TickTypes.ts
Vichy97: I think "send" only makes sense because "redpoint" and "pinkpoint" don't exist for boulders. I agree "send" is redundant for roped climbs and I don't plan to display it as an option in the mobile app, except for boulders.

Vichy97: > I don't want to be the annoying opinionated guy (especially as a new contributor)

I prefer people to be opinionated! I don't want to make these decisions siloed.

Vichy97: If you would like to keep "RedPoint" in addition to "Send" as a valid option I'm definitely ok with that!

Vichy97:
> Addressing: https://github.com/OpenBeta/openbeta-graphql/issues/419
@Vichy97 I simply made this a non-required field. Which is how I interpreted this issue.

Previously these fields were a non-required string, but I believe if you submit a tick and omit either of these fields, there was an error that was thrown. I don't see anywhere in your PR that changes this logic. Is there another part of the codebase that is doing validation around this?

prith-k:
> Previously these fields were a non-required string, but I believe if you submit a tick and omit either of these fields, there was an error that was thrown. I don't see anywhere in your PR that changes this logic. Is there another part of the codebase that is doing validation around this?
that would be here [src/db/TickSchema.ts](https://github.com/OpenBeta/openbeta-graphql/pull/444/files#diff-ff123a2d0d633abc7e42384cae4845e6a026b428bba900f4d210400d3ea3ba28)
`style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow'], required: false },
  attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Redpoint', 'Pinkpoint', 'Frenchfree', 'Fell/Hung', 'Send', 'Attempt'], required: false, index: true },`

Vichy97:
> > Previously these fields were a non-required string, but I believe if you submit a tick and omit either of these fields, there was an error that was thrown. I don't see anywhere in your PR that changes this logic. Is there another part of the codebase that is doing validation around this?
> > that would be here [src/db/TickSchema.ts](https://github.com/OpenBeta/openbeta-graphql/pull/444/files#diff-ff123a2d0d633abc7e42384cae4845e6a026b428bba900f4d210400d3ea3ba28)
> > `style: { type: Schema.Types.String, enum: ['Lead', 'Solo', 'TR', 'Follow'], required: false },   attemptType: { type: Schema.Types.String, enum: ['Onsight', 'Flash', 'Redpoint', 'Pinkpoint', 'Frenchfree', 'Fell/Hung', 'Send', 'Attempt'], required: false, index: true },`

Ah ok makes sense. I'm not super familiar with graphql server so I missed that! Thanks

src/db/TickTypes.ts
prith-k: The PR now reflects both send and redpoint in the attempt type. I personally support both as often on multi-pitch routes a send comes with an asterisk or two. And a redpoint often implies a multisession project.

Vichy97: Works for me!

glassbead0:
I started looking at the tickForm on the frontend. Is anyone else working on matching that with the code here? if not, I'll keep looking at it, and maybe combine it with with the MP import https://github.com/OpenBeta/open-tacos/pulls/1259
@Vichy97 @PrithipalKhalsa ?

Vichy97:
Nobody is working on it. I don't even see a way to view or add ticks on the front end

glassbead0:
Also, maybe this is a separate issue, but I noticd there is a uniqueness constraint on ticks: `E11000 duplicate key error collection: openbeta.ticks index: climbId_1_dateClimbed_1_style_1_userId_1_source_1`

So I cant have multiple ticks on the same route, same style, same day. that means I can't log 2 lead attempts on the same route on the same day, which i i should  be able to do.  I get that 2 "onsights" doesnt make sense, but i think the current index is too restrictive, and maybe should just be removed?

glassbead0:
> Nobody is working on it. I don't even see a way to view or add ticks on the front end

I see this on route pages: `We're giving this page a facelift! Visit the previous version]to make edits.` The 'previous version' does have a tick interface.... who has context on the 'new' vs 'old' route pages?

musoke:
> Also, maybe this is a separate issue, but I noticd there is a uniqueness
constraint on ticks: E11000 duplicate key error collection: openbeta.ticks
index: climbId_1_dateClimbed_1_style_1_userId_1_source_1
>
> So I cant have multiple ticks on the same route, same style, same day.
that means I can't log 2 lead attempts on the same route on the same day,
which i i should be able to do. I get that 2 "onsights" doesnt make sense,
but i think the current index is too restrictive, and maybe should just be
removed?

I _thought_ this was already relaxed.

glassbead0:
Here's a draft of the front-end "Tick This Route" button.  https://github.com/OpenBeta/open-tacos/pull/1260

It doesnt' exactly sync up with the code here. I'll take another look tomorrow.

vnugent:
> > Nobody is working on it. I don't even see a way to view or add ticks on the front end
> 
> I see this on route pages: `We're giving this page a facelift! Visit the previous version]to make edits.` The 'previous version' does have a tick interface.... who has context on the 'new' vs 'old' route pages?

The old climb page has inline editing (view and edit-mode are on the same page).  The code is also convoluted.  Last year we upgraded Next.js framework from v12 -> v13 and in the process, moving area editing into its own pages, simplifying the area view page.  The climb page hasn't received the same upgrade pattern until recently.  The new climb page is now view-only.  We're currently missing the climb-edit page(s).

src/model/TickDataSource.ts
```diff
@@ -74,6 +78,62 @@ export default class TickDataSource extends MongoDataSource<TickType> {
...
+    ['trad', 'sport', 'alpine', 'ice', 'mixed', 'aid'].some(
+      type => climb.type[type] === true)
+
+    const tickStyle = tick.style ?? 'null' // Provide a default value if tick.style is undefined
+    const attemptType = tick.attemptType ?? 'null' // Provide a default value if tick.attempy is undefined
+    if (isDWSOnly || isBoulderingOnly) { // bouldering and dws can only have attempt types: 'Send', 'Flash', 'Attempt', 'Onsight' and should have no sytle
+      if ((['Lead', 'Solo', 'Tr', 'Follow', 'Aid'].includes(tickStyle)) || ['Pinkpoint', 'Frenchfree', 'Redpoint'].includes(attemptType)) {
+        throw new Error('Invalid attempt type or style for DWS/Bouldering')
+      }
+    } else if (isTROnly) { // TopRope can only have attempt types: 'Send', 'Flash', 'Attempt', 'Onsight' and styles: 'TR'
```
glassbead0: This one doesn't quite make sense to me. 
1) TR only routes can also be Soloed (so we should allow 'Solo' as a type in addition to TR)
2) I thought we weren't going to allow 'Onsight' and 'Flash' for TR routes, and only allow 'Send'? I guess I can restrict that on the front end only, but allow those types in the DB.

src/graphql/schema/Tick.gql
```diff
@@ -120,6 +120,40 @@ enum TickSource {
...
+"""The type of attempt the user wants to tick for a route."""
+enum TickAttemptType {
+  "Onsight: Sending a route fist try with no prior knowledge"
+  Onsight
+  "Flash: Sending a route first try with prior knowledge"
+  Flash
+  "Pinkpoint: Sending a route with no falls on pre-placed gear"
+  Pinkpoint
+  "French Free: Doing the route with pulling on gear or draws"
+  FrenchFree
```
glassbead0: `Frenchfree` (instead of `FrenchFree`) to match above.

src/model/TickDataSource.ts
Vichy97: I was planning on not showing attempt types for TR in the UI. This is how MP works currently. "Solo" is already a style.

prith-k: I don't have a strong opinion about this one. But people do use the term 'trash' as a portmanteau of TR and flash, so the concept of flashing on toprope is used.

src/graphql/schema/Tick.gql
```diff
@@ -120,6 +120,40 @@ enum TickSource {
...
+  Onsight
+  "Flash: Sending a route first try with prior knowledge"
+  Flash
+  "Pinkpoint: Sending a route with no falls on pre-placed gear"
+  Pinkpoint
+  "French Free: Doing the route with pulling on gear or draws"
+  Frenchfree
+  "Attempt"
+  Attempt
+  "Send: Sending a route with no falls"
```
Vichy97, line 135: I'm realizing we need "fell/hung" as a separate value. I was originally planning on treating "attempt" as "fell/hung" for roped climbs, but we don't always have a way to tell if the climb was a roped climbed or a boulder. ie: if the tick was imported from MP but doesn't have a corresponding entry in our DB. I think we should add back "fell/hung" to support this

glassbead0: After discussion at sprint planning, we are NOT going to use "Fell/Hung". I've updated the logic in the MP import code https://github.com/OpenBeta/open-tacos/pull/1259 to handle this, and I've written up a markdown document about tick logic, since it's complicated. Also, we decided to store 'Boulder' in the 'style' field (unlike mountain project), which i think will make our data look much cleaner. 

 @PrithipalKhalsa   I had to make a couple changes to your code to get MP imports working along with PR 1259. Would you mind cherry-picking this commit ~https://github.com/glassbead0/openbeta-graphql/commit/1341fd5fdfa2962236440a8bc884edb798138a6c~ https://github.com/glassbead0/openbeta-graphql/commit/0aca80a944d93154e0bb2f9d3e5a2a9aa9d4140c
into this PR? Thanks! (that commit also has the markdown doc)

vnugent:
@all-contributors add @PrithipalKhalsa for code and ideas

Issue #442, opened 2025-01-03
Enable batched queries


Vichy97:
Batched queries will be vital for supporting offline mode for the mobile app. It's not realistic to do one large query as this can take upwards of a full minute for an entire state and often times out or fails. Apollo makes it super easy to split this into batches, but only if it's supported on the backend.

Vichy97:
https://www.apollographql.com/docs/graphos/routing/performance/query-batching This has all the documentation including sample queries. It looks like it's just a couple lines to configure this feature.

glassbead0:
I'm currently working on this. I'm new to the project, so it will likely take me a bit longer, and if anyone wants to join, ping me on discord (Infinity00)

@Vichy97, since I'm new to openbeta, can you provide more specific requirements for this? If you could give me an example of the one large query you're currently making, and what kind of batched query you want to make? Happy to chat in Discord too.

Vichy97:
I would like the ability to query all the climbs and areas in an entire state. I plan to have that query automatically broken down and sent as batched queries using the Kotlin Apollo client. I believe this is basically automatically handled for you on the server side as long as you follow the configuration in that documentation I linked.

vnugent:
@Vichy97 according to the page you sent the batching feature is only available to Enterprise plan.
@glassbead0 once you get your development env working, I can find you an simpler task to get started.

In general the slowness you're seeing is a classic n+1 problem.  Consider this query, getting info & all images for Red Rock and the next immediate children and their images.

```gql
query MyQuery {
  area(uuid: "bea6bf11-de53-5046-a5b4-b89217b7e9bc") {
    areaName
    media {
      id
    }
    children {
      areaName
      media {
        id
      }
    }
  }
}
```

For each child of Red Rock (let's say there's 20 of them), the backend sends a separate query to the media collection, or 20 queries.  We're subclassing the mongo data source class that is supposed to be smart enough to run only 1 query,
 eg select * from collection where id **in** ( set of child IDs), vs 20 individual  "select * from media where id = child_id1".  

We'll need to rewrite how we resolve the media field or add a new batch download query for mobile to download multiple areas + photos at once.

vnugent:
See PR https://github.com/OpenBeta/openbeta-graphql/pull/451

PR #451, opened 2025-02-26
bulk areas query

vnugent:
Issue #442 

- New query `bulkAreas(ancestors: string[]): AreaType[]` to get all areas with a common ancestor path.   **ancestors** is an array of area uuid's starting from the country root node, which is same as an area `ancestors` field.
- Max payload is [hard coded at 10MB](https://github.com/OpenBeta/openbeta-graphql/blob/247f8a95b72dc78ed4bce39b1270b38c8e116bbc/src/server.ts#L61).  I haven't tested querying some of the largest states (Colorado, CA, Utah) with all the usual fields.

In this example, we're getting Arizona and its descendants, another word any areas with ancestors starting with USA / Arizona.

Screenshot 2025-02-26 at 6 07 23 PM

@Vichy97 would something like this work?

### DIFF
--- a/src/db/AreaTypes.ts
+++ b/src/db/AreaTypes.ts
@@ -257,3 +257,7 @@ export enum OperationType {
   /** Set areas' sorting index */
   orderAreas = 'orderArea'
 }
+
+export interface BulkAreasGQLQueryInput {
+  ancestors: string[]
+}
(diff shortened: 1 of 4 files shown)


Vichy97:
This would work, but is there a reason we can't just pass an area id?

vnugent:
> This would work, but is there a reason we can't just pass an area id?

It's a small performance optimization (one less look up query).  I can look up ancestors from area id, but the caller, the mobile app in this case, most likely has ancestors.

vnugent:
Also, to reduce misuse the call will require the user's API key similar to update mutations.

Vichy97:
Ok sounds good. I'll test this out when it lands

PR #453, opened 2025-02-28
Bulk adding new routes - ordering of routes gets scrambled and requires reordering

greghart:
- [x] We have a test coverage for adding climbs, but also add test for how new climb gets ordered with existing climbs

---
name: Pull request
about: Create a pull request
title: Bulk adding new routes - ordering of routes gets scrambled and requires reordering
labels: bug, feature ?
assignees: Greg Hart

---

## What type of PR is this?(check all applicable)
- [ ] refactor
- [x] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description

Per related issue, when adding multiple climbs, we don't persist the ordering 

### Related Issues

Issue [#1255](https://github.com/OpenBeta/open-tacos/issues/1255)

### What this PR achieves

This PR changes `addOrUpdateClimbs` to persist the order of incoming arrays of climbs using `left_right_index` (when not already given0.
It also formalizes whether new climbs should come after or before existing climbs (when `left_right_index` is not given) -- currently it changes this so new climbs come after instead of before.

### Screenshots, recordings

image

### DIFF
--- a/src/model/MutableClimbDataSource.ts
+++ b/src/model/MutableClimbDataSource.ts
@@ -110,16 +110,17 @@ export default class MutableClimbDataSource extends ClimbDataSource {
     // when adding new climbs, ensure they are added in expected sort order, such that:
     // * multiple climbs keep the order they're input in.
     // * all newly added climbs come after existing climbs.
-    let newClimbLeftRightIndex = (await this.climbModel
+    const maxExisting = (await this.climbModel
       .find({ _id: { $in: parent.climbs } })
       .select('metadata.left_right_index')
       .sort({ 'metadata.left_right_index': -1 })
-      .limit(1))[0]
-      ?.metadata.left_right_index ?? 1
+      .limit(1))[0]?.metadata.left_right_index
+    let newClimbLeftRightIndex = (maxExisting ?? 0) + 1
+
     function resolveLeftRightIndex (i: number): { left_right_index: number } | null {
       // user input is always prioritized
       if (userInput[i].leftRightIndex !== undefined) {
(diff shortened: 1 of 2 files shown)


src/model/MutableClimbDataSource.ts
```diff
@@ -107,6 +107,28 @@ export default class MutableClimbDataSource extends ClimbDataSource {
...
+    const maxExisting = (await this.climbModel
+      .find({ _id: { $in: parent.climbs } })
+      .select('metadata.left_right_index')
+      .sort({ 'metadata.left_right_index': -1 })
+      .limit(1))[0]?.metadata.left_right_index
+    let newClimbLeftRightIndex = (maxExisting ?? 0) + 1
+
+    function resolveLeftRightIndex (i: number): { left_right_index: number } | null {
+      // user input is always prioritized
+      if (userInput[i].leftRightIndex !== undefined) {
```
vnugent: can you use `!=` operator to account for nullness?

vnugent (changes requested):

src/model/MutableClimbDataSource.ts
greghart: No problem, resolved ✔️ 
Missed this convention, have historically used eslint which prefers triple equal (https://eslint.org/docs/latest/rules/eqeqeq).

I guess in this case, while the type declaration `leftRightIndex?: number` tells us it'll be a `number | undefined`, we don't actually validate that since data comes in as `any` from graphql, so client is free to send `null` and we should handle that as well

vnugent:
@all-contributors add @greghart for code & ideas

PR #455, opened 2025-03-14
fix: allow water ice grades

glassbead0:
Companion PR to https://github.com/OpenBeta/open-tacos/pull/1282


src/db/ClimbTypes.ts
```diff
@@ -100,15 +100,6 @@ export enum SafetyType {
   X = 'X',
 }
 
-export interface IGradeType {
```
glassbead0, line 103: this appears to be unused.

Issue #458, opened 2025-04-02
improve speed of BulkArea query


glassbead0:
Querying BulkArea for an entire state is slow. this is because of the N+1 problem when querying for media, climbs, and organizations for each area. To improve perfomance, we should update the code to use a JOIN query to get all the data (areas, climbs, media, organizations), in one query.

CocoisBuggy:
This issue could be partially resolved with #428 since it brings alternate parent/child representations that wouldn't explode the backend, so I would be happy to resolve this if that gets merged

PR #459, opened 2025-04-02
feat: return all leaf+non-leaf areas

### LINKED ISSUE #457: BulkArea query only returns non-leaf areas
Perform a BulkArea query with ancestors:
"1db1e8ba-a40e-587c-88a4-64f5ea814b8e",
        "24e148bb-4884-59e8-9870-43ce21f1a337",
        "c1fdee82-93e7-5a55-9103-7defbe7f0b10",
        "acd5e4e7-2972-5435-a237-4cea3dffe6cb",
        "f84f887c-a079-5668-ac10-8ee676aa5f8a"

this should return 5 results, but it only returns 1.

This is because the other 4 areas are leaf areas, and [this function](https://github.com/OpenBeta/openbeta-graphql/blob/develop/src/model/AreaDataSource.ts#L185) defaults to not returning leaf areas.

We should return both leaf and non-leaf areas.

glassbead0:
Resolves [this issue](https://github.com/OpenBeta/openbeta-graphql/issues/457)

BulkAreas query now return both leaf and non-leaf areas, instead of only non-leaf areas. It also fetches leaf and non-leaf areas in paralles, so this hopefully won't be much or any of a performance hit on the bulkAreas query.

### DIFF
--- a/src/model/AreaDataSource.ts
+++ b/src/model/AreaDataSource.ts
@@ -287,6 +287,10 @@ export default class AreaDataSource extends MongoDataSource<AreaType> {
       })
     }
     const ancestorsCSV = ancestors.join(',')
-    return await this.findDescendantsByPath(ancestorsCSV)
+    const [leafAreas, nonLeafAreas] = await Promise.all([
+      this.findDescendantsByPath(ancestorsCSV, true),
+      this.findDescendantsByPath(ancestorsCSV, false)
+    ])
+    return nonLeafAreas.concat(leafAreas)
   }
 }


glassbead0:
@vnugent thoughts on merging this to get Vincent unblocked on getting all the correct areas returned? 

I'll work the larger performance improvement issue separately.

src/model/AreaDataSource.ts
```diff
@@ -287,6 +287,10 @@ export default class AreaDataSource extends MongoDataSource<AreaType> {
       })
     }
     const ancestorsCSV = ancestors.join(',')
-    return await this.findDescendantsByPath(ancestorsCSV)
+    const [leafAreas, nonLeafAreas] = await Promise.all([
+      this.findDescendantsByPath(ancestorsCSV, true),
+      this.findDescendantsByPath(ancestorsCSV, false)
+    ])
```
Vichy97, line 293: This works, but couldn't you just remove the boolean from the method entirely? When would you only want leaf vs non-leaf?

Vichy97 (approved):

src/model/AreaDataSource.ts
glassbead0: I could do it that way, but this way does have a little bit of a performance benefit, depending on the ratio of leaf to non-leaf areas.

I will say, locally leaf areas account for ~90% of all areas in Colorado and California (which i tested). So I'm concerned that you'll see a huge slow down after this merge, but at least for now, better to get the right data slowly, than get not nearly enough data.

I'm going to go ahead and merge, and i'll do my best to work on [speeding up performance properly](https://github.com/OpenBeta/openbeta-graphql/issues/458) in the next couple weeks.

Issue #443, opened 2025-01-08
Add an api to upload an image


Vichy97:
Right now I need to do a lot of steps to upload from mobile. I need to:
- generate a uuid
- request a url from google url service with that uuid in the url
- upload the blob to that url
- make a mutation to include the tag for the area/climb i'm uploading to

This is a lot of steps and not very robust in the case of failure of any of the steps. It also will be very difficult to support uploading images while offline. We should have a single mutation that does all of this.

CocoisBuggy:
I wouldn't mind cleaning up some of the flow to do with the media upload. @Vichy97 are you hoping to pass the blob through the gql layer and then have it resolve as an openbeta media object, or are you just wanting to see some of the steps reduced / the control flow made a bit more clear?

Vichy97:
Ya I'd love to be able to just upload the blob and get the media object back. I don't think the client should have to generate the uuid and urls or have to make multiple api calls.

CocoisBuggy:
References:

```mermaid
sequenceDiagram
    User->>Openbeta: Can I have a signed URL for uploading a file?
    Openbeta->>Openbeta: Check user auth
    Openbeta->>Google Cloud: Can I have a signed url?
    Google Cloud -->> Openbeta: Here is a signed url
    Openbeta-->>User: you are authenticated, so here is your url
    User->>Google Cloud: Here is the file
    Google Cloud-->>User: Okay
    User ->> Openbeta GQL: Okay, the media is created, here is the url.
    Openbeta GQL-->>User: Okay
```

This is a lot of moving parts holding their own state that may become orphaned if the user drops out at some point in the process. The suggested flow is something like

```mermaid
sequenceDiagram
    User->>Openbeta GQL: Here is an image file
    Openbeta GQL->>Openbeta GQL: Check user auth
    Openbeta GQL->>Google Cloud: Here is a file for you
    Google Cloud -->> Openbeta GQL: Okay
    Openbeta GQL->> Openbeta GQL: Create entry in database
    Openbeta GQL -->>User: Here is the media object
```

This cuts down on the moving pieces quite significantly, as now the user only needs to make one request, but it is by no means a free lunch. The most viscous issue being that file upload is very heavy for our poor little GQL endpoint. In a kinder world our little javascript baby would be better at concurrent IO, but as it stands nodejs humbles our api a little bit. However, do not fret. There is a nearly-as-convenient construct we can use, which looks like this.

 ```mermaid
sequenceDiagram
[...]

PR #465, opened 2025-04-12
Media Upload Tools

CocoisBuggy:
Working a little bit on #443, and this is where my thinking is taking me so far.

There isn't consensus on this being the best approach - the alternate is discussed in the issue comment chain - but I wanted to take a look and make sure my proposal was feasible

I have written code and tests for both subscription paradigms that google supports, but I guess realistically we need to choose one and curb the other one. @vnugent I'm not really an expert in GCS, but as I see it the choice is down to how we do load balancing.
My initial intuition was that we should just do a web hook since then the requests from google will flow through whatever balancing strategy is in place, [but I see some evidence](https://cloud.google.com/pubsub/docs/subscriber) that actually the subscribers get a subset of the messages and that some load balancing is already in place. Do you have any idea? We can go either way since both options are implemented and tested

Next step is just to update the gql mutation and add some tests and then the feature should be done. We can roll it out without affecting any of the existing media upload features in opentacos.

I've put some extra documentation in the relevant README [here](https://github.com/OpenBeta/openbeta-graphql/blob/feat/%23443/src/google-cloud/README.md)


CocoisBuggy:
image
Fortunately for everyone it looks like I can simplify this PR to JUST use the pull subscriptions, which will make life more pleasant for everyone

image
Also confirms that subscriptions can be shared horizontally

PR #469, opened 2025-04-24
feat:return ancestors and pathTokens on climbs in area queries

glassbead0:


glassbead0:
Full transparency: I used GH copilot for the changes in the BulkArea query. It works, and the climbs now contain ancestors and pathTokens from their parent area. This obviously increases the overall size of the response (in my testing, by about 50%!!), and thus also increases the time the query takes by about 50%.

PR #474, opened 2025-05-12
Feat: Allow for query of cumulative image weight

### LINKED ISSUE #417: Areas should include image size metadata
For the mobile client it would be great to let the user know how large a download will be before they confirm to download. Exposing metadata with the media size total would make this much easier.

CocoisBuggy:
This was requested by our stalwart mobile dev, and it does make a sort of sense - though I'm not 100% sure the metric will be totally accurate
- it might be worthwhile making another issue for the actual TEXT weight on top of the media byte-sizes all summed together

resolves #417

### DIFF
--- a/src/__tests__/areas.ts
+++ b/src/__tests__/areas.ts
@@ -98,6 +98,27 @@ describe('areas API', () => {
         })
     })
 
+    it('retrieves an area and its cumulative media weight', async () => {
+      const response = await queryAPI({
+        query: `
+          query area($input: ID) {
+            area(uuid: $input) {
+              uuid
+              imageByteSum
+            }
+          }
+        `,
+        operationName: 'area',
+        variables: { input: ca.metadata.area_id },
+        userUuid,
+        app
+      })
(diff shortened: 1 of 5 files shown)


glassbead0:
I tested this code locally, and it was quite slow to request the imageByteSum for the USA (21 seconds, with 81 MB as the result). it was much faster at the state level (<1 second with as much as 50 MB), so I think just getting all descendants at the country level is quite slow. I imagine this will mostly be used at the state level though, so probably not really a big deal. 

This looks good to me!

glassbead0 (approved):

CocoisBuggy:
Thanks for bench-marking it! yea it is brutally slow, but my hope is that users will have no reason to fetch this value unless they are doing cache lookahead, which shouldn't be every time (I hope)

It could be sped up enormously by de-normalizing the field but the complexity of that approach is very much non-zero.

The actual way to speed it up is to merge #428 which has a more performant implementation of our area structure and would eliminate the expensive regex needed in this approach

PR #477, opened 2025-05-14
Update nodejs to latest LTS

CocoisBuggy:
A noisy one for sure. I strongly feel this will be necessary for us since node 18 is end of life and v22 is the next one on the list.

 - Update node
 - update typescript
 - remove ts-standard and move to eslint
 - propogate compliance (hence the noise, since some of our files were ignored by the previous lint system)

I'd like to slowly enable many of the rules that I have suppressed in the linter config, since many of them represent modern best practice but I will make issues for each one individually so as not to dump too many changes here


CocoisBuggy:
Ah yes, that was always going to fail in CI. woops

CocoisBuggy:
@vnugent will need to comment on if there are any ramifications for production that I need to account for

